PR #206, opened 2020-06-10
Restore continuous integration

caguero:
After the migration from BitBucket to GitHub we lost CI. This PR migrates CI from `BitBucket` Pipelines to `GitHub` workflows using actions.


bsb808:
I'll take a look.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+        shell: bash
+        run: |
+          cd /tmp/vrx_ws;
+          source /opt/ros/melodic/setup.bash;
+          catkin_make -j1;
+          catkin_make run_tests -j1;
+
+  test-gazebo7-kinetic:
+
+    runs-on: ubuntu-18.04
```
bsb808, line 37: Just curious really - should this run on 16.04?

caguero: I might be wrong because I learned all I know about GitHub workflows and actions yesterday but I think this is the Ubuntu distribution that the GitHub runner (the VM) is using. On top of that, we choose to run our own Docker container (which is 16.04 in this case).

Issue #69, opened 2019-03-01
Configure osrf/vrx Dockerhub repository


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/vrx-gh-pages/#!/osrf/vrx/issues/69)) by Michael McCarrin (Bitbucket: [m1chaelm](https://bitbucket.org/%7B3c3f500d-bbf6-472c-b8f7-627e5eabc226%7D/)).**

----------------------------------------
New repository for new name (vrx) needs to be set up so it's linked to the bitbucket code. Content we want to keep from the old Dockerhub repo should be ported over. Autobuild should be set up to produce images when new releases are published. Since release 1.0 is already published, this will need to be built and pushed manually.

caguero:
See https://github.com/osrf/vrx/pull/208

M1chaelM:
Resolved with #208 but never closed. Now moot anyway because we no longer release Docker images with VRX compiled. This has been replaced with base images containing the correct dependencies, which are available in the [dockwater](https://github.com/Field-Robotics-Lab/dockwater) repository.

PR #208, opened 2020-06-18
Automatic release of Docker images

caguero:
This pull request should fix issue #69 . The idea is that when a new release tag is created, we build a Docker image and push it to DockerHub automatically. A release tag is something that starts with `v` followed by the version number (e.g.: `v1.3.1`).

This GitHub workflow™ builds the Nvidia and non-Nvidia versions of the Docker image that uses Ubuntu Bionic/ROS Melodic/Gazebo9.

I did some testing and the images were correctly uploaded to https://hub.docker.com/r/osrf/vrx/tags

### DIFF
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -2,7 +2,7 @@ name: Release to DockerHub
 on:
   push:
     tags:
-      - v*  # Launch DockerHub push on v* tags (e.g.: v1.2.1)
+      - 'v*'  # Launch DockerHub push on v* tags (e.g.: v1.2.1)
   
 jobs:
   release-docker-image-nonvidia-bionic-melodic-gazebo9:
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -32,6 +32,8 @@ jobs:
 
       - name: Create a new VRX Docker image
         run: |
+          sudo apt update;
+          sudo apt-get install -y docker;
           cd docker;
           ls;
           bash +xe ./build.bash ${{env.build_docker_flags}} ..;
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -19,13 +19,24 @@ jobs:
         if: ${{ success() }}
         run: |
           sudo apt update;
-          sudo apt-get install -y docker
+          sudo apt-get install -y apt-transport-https \
+                                  ca-certificates \
+                                  curl \
+                                  gnupg-agent \
+                                  software-properties-common;
+          curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+          sudo add-apt-repository \
+            "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
+            $(lsb_release -cs) \
+            stable";
+          sudo apt-get update;
+          sudo apt-get install docker-ce docker-ce-cli containerd.io;
 
       - name: Run build script
         shell: bash
         if: ${{ success() }}
         run: |
-          ls;
+          ls -a;
           docker/build.bash .
 
       - name: Tag image
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -1,8 +1,9 @@
 name: Release to DockerHub
-on:
-  push:
-    tags:
-      - v*  # Launch DockerHub push on v* tags (e.g.: v1.2.1)
+# on:
+#   push:
+#     tags:
+#       - v*  # Launch DockerHub push on v* tags (e.g.: v1.2.1)
+on: [pull-request]
   
 jobs:
   release-docker-image-nonvidia-bionic-melodic-gazebo9:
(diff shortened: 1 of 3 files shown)


caguero:
@M1chaelM , maybe you can take a look.

M1chaelM:
Carlos,

Apologies for the delay. I had another project come up with a fairly short fuse. I should be able to get to this next week.

Michael

caguero:
> @caguero ,
> 
> Apologies for the delay. I had another project come up with a fairly short fuse. I should be able to get to this next week.
> 
> Michael

Sounds good, thanks!

docker/Dockerfile
```diff
@@ -70,10 +70,9 @@ RUN /bin/sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc)
     ros-${DIST}-velodyne-simulator \
     ros-${DIST}-rqt \
     ros-${DIST}-rqt-common-plugins \
- && rosdep init \
  && apt clean
 
-RUN rosdep update
+RUN /bin/bash -c ". /opt/ros/${DIST}/setup.bash"
```
M1chaelM, line 75: I noticed rosdep is no longer installed by default so `rosdep init` and `rosdep update` throw errors. The [ROS Melodic Installation Instructions](http://wiki.ros.org/melodic/Installation/Ubuntu) resolve this by installing `python-rosdep` explicitly. I'm assuming this edit indicates we don't really need to do this--would you mind explaining why?

M1chaelM: @caguero Just bumping this. I'm confused because it looks like running setup.bash replaces the calls to rosdep, but setup.bash seems just to set environment variables, and since it is on it own line I expect these to disappear before they are used. Can you say what's going on here?

caguero: Oh, if I remember correctly, I saw this error:

```
/bin/sh: 1: rosdep: not found
```

And I removed these two lines as a quick test but I forgot to restore it. I think the proper way to do it is, as you suggested, to install `python-rosdep`.

caguero: It should be restored now. By the way, I had to change the way Docker is installed. The old way doesn't work anymore (maybe something changed in Ubuntu) and now I'm using another Github action to do it.

M1chaelM: Thanks! Good to know about the new Docker install commands.

I think line 79 can be eliminated. It appears to be setting environment variables, and these don't persist between RUN commands in Docker files. I ran a quick test by dropping it from my local file and the image still appears to build fine and runs vrx.launch without error.

caguero: Thanks! Removed in d356499.

caguero:
@M1chaelM , any more suggestions or are you ready to approve?

.github/workflows/docker_release.yml
```diff
@@ -0,0 +1,116 @@
+name: Release to DockerHub
+on:
+  push:
+    tags:
+      - v*  # Launch DockerHub push on v* tags (e.g.: v1.2.1)
```
M1chaelM, line 5: I believe the v* needs to be in quotations, i.e. 'v*'
Without those quotations I was unable to get this action to trigger.

M1chaelM (changes requested):
Sorry for the delay. This was difficult to test. I forked our repository to my personal github space, set up a corresponding Docker Hub repository and modified the action to point to it instead of osrf/vrx. Then I created a bunch of fake tags on the fork to make sure the trigger was working, which lead to the discovery in my comment that quotations are needed around the v* expression.

I think with this modification the PR is ready to be submitted. I attempted to make the modification myself but for some reason I am unable to push to the repository. 

That said, I see also that the branch is now out of date with the base and is not passing the CI checks for some reason, so this may need to be fixed also.

.github/workflows/docker_release.yml
caguero: Thanks for the quotes! :+1:

caguero:
> Sorry for the delay. This was difficult to test. I forked our repository to my personal github space, set up a corresponding Docker Hub repository and modified the action to point to it instead of osrf/vrx. Then I created a bunch of fake tags on the fork to make sure the trigger was working, which lead to the discovery in my comment that quotations are needed around the v* expression.
> 
> I think with this modification the PR is ready to be submitted. I attempted to make the modification myself but for some reason I am unable to push to the repository.
> 
> That said, I see also that the branch is now out of date with the base and is not passing the CI checks for some reason, so this may need to be fixed also.

I think we're good now. Let me know if you see any other issues.

M1chaelM:
> > Sorry for the delay. This was difficult to test. I forked our repository to my personal github space, set up a corresponding Docker Hub repository and modified the action to point to it instead of osrf/vrx. Then I created a bunch of fake tags on the fork to make sure the trigger was working, which lead to the discovery in my comment that quotations are needed around the v* expression.
> > I think with this modification the PR is ready to be submitted. I attempted to make the modification myself but for some reason I am unable to push to the repository.
> > That said, I see also that the branch is now out of date with the base and is not passing the CI checks for some reason, so this may need to be fixed also.
> 
> I think we're good now. Let me know if you see any other issues.

Ugh, sorry--I requested this change when I didn't have write access, then fixed it myself once I did, but I didn't realize I have to take additional steps to acknowledge that changes are no longer being requested. Trying to figure out how to do this...

M1chaelM (approved):
Approved!

PR #214, opened 2020-09-15
New hydro params

bsb808:
Exposes additional hydrodynamic parameters, specifically added mass and quadratic drag parameters.  These are used by the USV model in the Frontiers paper.  Also used for a non-VRX USV model.

To test, simply run the standard example...
```
```

Nothing should change since the current WAM-V model doesn't specify these parameters so the default, nominal values are used by they hydro model.


caguero (approved):
I added some documentation and the changelog in c04688f. Other than that looks good to me!

PR #219, opened 2020-10-03
Use custom wind topics

caguero:
These debug topics were hardcoded. Now we can change them if needed:

```
/vrx/debug/wind/speed
/vrx/debug/wind/direction
```


mabelzhang (approved):
Worked for me. Tested in https://github.com/osrf/vorc/pull/11

PR #226, opened 2020-10-30
Set CMake 3.1+ policies explicitly to remove warnings

mabelzhang:
To support adding CI for VORC https://github.com/osrf/vorc/pull/21

CMake policies [CMP0046](https://cmake.org/cmake/help/v3.0/policy/CMP0046.html), [CMP0054](https://cmake.org/cmake/help/v3.1/policy/CMP0054.html), and [CMP0071](https://cmake.org/cmake/help/git-stage/policy/CMP0071.html) started introducing warnings in CMake 3.10.2, 3.1.3, and 3.10, respectively.

These produce warnings when using newer CMake versions.

To reproduce the warnings, on Ubuntu 18.04 (target platform for ROS Melodic, `apt` ships CMake 3.10.2), compile from scratch:
```
$ rm -rf build devel log install
$ catkin_make
```

This PR removes those warnings properly.

`OLD` and `NEW` flags are set explicitly to appease CMake warnings.
Specifically, CMP0054 is for using variables in quotes in `FindBoost.cmake`, which we cannot fix.
CMP0071 is for using `AUTOMOC` and `AUTOUIC` in `vrx_gazebo` with generated files in `vrx_gazebo/msgs` (from protobuf).
CMP0046 is for building messages and services in the same package as where they're used, by `add_dependencies(*generate_messages_cpp)` and `generate_services_cpp`, packages that do not exist yet until after everything is built.

### DIFF
--- a/usv_gazebo_plugins/CMakeLists.txt
+++ b/usv_gazebo_plugins/CMakeLists.txt
@@ -1,5 +1,12 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(usv_gazebo_plugins)
+
+# Set policy for CMake 3.1+, FindBoost.cmake dependency of gazebo uses quoted
+# variables in if()
+if(POLICY CMP0054)
+  cmake_policy(SET CMP0054 OLD)
+endif()
+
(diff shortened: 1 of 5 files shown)


caguero (approved):
Looks good!

Issue #237, opened 2021-02-10
Add a noetic branch.


rolker:
Can #230 be accepted into a new noetic branch?

rolker:
On second thought, the difference is just a small change in the CMakeList.txt files so maybe detecting the distro and adjusting in the cmake files is better than multiple branches. I attempted such a fix (#239) and used a PR to trigger the CI to test it for me on kinetic since I don't have an easy way to do so myself, but as of now it failed.

If anyone has hints on how to solve this, I'm all ears!

Issue #236, opened 2021-02-10
Gazebo crashes when loading a model with usv_gazebo_dynamics_plugin in default world without a wave model


rolker:
A null waveParams pointer used when calling WavefieldSampler::ComputeDepthSimply causes a crash.

I have a simple fix where I skip the call if waveParams is null and use a 0 value.

I'll send the PR, but since I'm developing on noetic, it also includes #230.

PR #230, opened 2020-11-25
WIP: Fix build failure for ROS Noetic

d130s:
Same as https://github.com/osrf/vorc/pull/32.

For the same reasons, **DO NOT MERGE** unless the targetted ROS version is changed to Noetic or later. Hene, this is made as WIP.

PR #238, opened 2021-02-10
Allow null wave params

rolker:
Fix for #236.

It was developed on noetic so it includes #230. It should probably not be merged into master. If #237 gets done, I can resubmit against a noetic branch.

### DIFF
--- a/usv_gazebo_plugins/src/usv_gazebo_dynamics_plugin.cc
+++ b/usv_gazebo_plugins/src/usv_gazebo_dynamics_plugin.cc
@@ -330,8 +330,9 @@ void UsvDynamicsPlugin::Update()
       //  *waveParams, X, simTime);
       double depth = 0.0;
       if (waveParams)
-        depth = WavefieldSampler::ComputeDepthSimply(
-        *waveParams, X, simTime);
+      {
+        depth = WavefieldSampler::ComputeDepthSimply(*waveParams, X, simTime);
+      }
 
       // Vertical wave displacement.
       double dz = depth + X.Z();
--- a/usv_gazebo_plugins/src/usv_gazebo_dynamics_plugin.cc
+++ b/usv_gazebo_plugins/src/usv_gazebo_dynamics_plugin.cc
@@ -329,7 +329,7 @@ void UsvDynamicsPlugin::Update()
       // double depth = WavefieldSampler::ComputeDepthDirectly(
       //  *waveParams, X, simTime);
       double depth = 0.0;
-      if(waveParams)
+      if (waveParams)
         depth = WavefieldSampler::ComputeDepthSimply(
         *waveParams, X, simTime);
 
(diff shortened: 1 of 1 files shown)


usv_gazebo_plugins/src/usv_gazebo_dynamics_plugin.cc
```diff
@@ -328,7 +328,9 @@ void UsvDynamicsPlugin::Update()
       double simTime = kTimeNow.Double();
       // double depth = WavefieldSampler::ComputeDepthDirectly(
       //  *waveParams, X, simTime);
-      double depth = WavefieldSampler::ComputeDepthSimply(
+      double depth = 0.0;
+      if (waveParams)
```
mabelzhang, line 332: Does the check on https://github.com/osrf/vrx/blob/master/usv_gazebo_plugins/src/usv_gazebo_dynamics_plugin.cc#L182-L188 not give it a non-null value?

rolker: In one of my use cases, it does not since that world does not have a wave model.

While developing the USV model, I often load it in an empty gazebo world, which worked fine until I added the dynamics plugin which led to a crash without many clues to what happened.

Now I understand that the plugin is meant to be used in a world with waves, but I think that gazebo shouldn't crash in the case the wave model is missing. Perhaps a warning or error message might be better, but this is just a quick fix to prevent the crash.

mabelzhang: Yeah, I think it is meant to be loaded with waves. I also see your point that the code should be more robust and shouldn't crash on a null pointer.
Just curious, do you get that message on line 184 printed over and over again?

rolker: No, I don't get that message at all. I tried with and without the debug flag on when launching gazebo.

rolker: Update: I just found the verbose flag and that message does continuously print with it on.

mabelzhang: Oh yeah that's right, it won't show up in normal mode.

mabelzhang (approved):
Thanks for the patch for robustness!

usv_gazebo_plugins/src/usv_gazebo_dynamics_plugin.cc
```diff
@@ -328,7 +328,9 @@ void UsvDynamicsPlugin::Update()
...
+        depth = WavefieldSampler::ComputeDepthSimply(
         *waveParams, X, simTime);
```
mabelzhang, line 334: Actually I'm gonna be nitpicky here and suggest style changes
```suggestion
      if (waveParams)
      {
        depth = WavefieldSampler::ComputeDepthSimply(*waveParams, X, simTime);
      }
```

PR #239, opened 2021-02-10
Support kinetic melodic and  noetic

rolker:
An alternative to creating a noetic branch (#237 ) to solve #230  is to check in the CMakeLists.txt files for the ROS distribution to determine the need for INLINE.

### DIFF
--- a/usv_gazebo_plugins/CMakeLists.txt
+++ b/usv_gazebo_plugins/CMakeLists.txt
@@ -104,17 +104,17 @@ install(TARGETS usv_gazebo_acoustic_pinger_plugin
   RUNTIME DESTINATION bin
 )
 
-set(XACRO_INLINE)
+set(XACRO_INORDER)
 if(DEFINED ENV{ROS_DISTRO})
   if($ENV{ROS_DISTRO} STREQUAL "kinetic")
-    set(XACRO_INLINE INLINE)
+    set(XACRO_INORDER INORDER)
   endif()
 endif()
 
 # Generate demo world files from xacro and install
 xacro_add_files(
   worlds/buoyancy_plugin_demo.world.xacro
-  ${XACRO_INLINE} INSTALL DESTINATION worlds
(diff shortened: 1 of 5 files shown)


d130s (approved):
Thanks for taking over #230 with commits being preserved!

rolker:
No problem @d130s , wanted to make sure you got credit for your part.

Unfortunately, my fix doesn't seem to have passed CI on kinetic and I don't have a kinetic setup to test and debug, so I'm hoping someone out there might be able to help?

mabelzhang:
The Kinetic failure looks like an error on the new addition on the `INLINE`. I think it's a typo that's supposed to say `INORDER`?

rolker:
Oh yes! That makes sense @mabelzhang. I'll push a fix and see if it passes CI.

mabelzhang:
FYI this repo is currently "maintained" and not in active development, so response time can be slow.
This one looks quick enough. If it's something involved, I can't guarantee we'll be testing and merging PRs right now, just so you know.

rolker:
Thanks for the heads up regarding maintenance status. 

I'm using the WAM-V model as well as the Cora model from VORC as guides to building a gazebo simulation of our own USV.

Our model does use the usv_gazebo_plugins package and according to http://wiki.ros.org/usv_gazebo_plugins, this seems to be where I should contribute PRs and such.

I'll be working from my own fork anyway so I don't mind if PRs here sit around for a while. Should I keep sending some anyway? I don't expect many since the innards of the plugins probably won't be where I'm focusing my efforts.

mabelzhang:
Glad it's of help to you!

Of course, feel free to send PRs, it could help others even if we aren't getting around to them. Thank you!

mabelzhang (approved):
CI is passing. Thank you for the fix to enable all three distros!
We haven't run on Noetic, but since this works for you both, I'll merge it. We don't have CI set up for Noetic at the moment, so I'd say Noetic is community-supported right now.

PR #240, opened 2021-02-23
Update GPS datum to more closely match nautical chart

rolker:
Allows the use of nautical charts and other real-world geo-referenced maps for mission planning and monitoring.

On a related note, the hector gazebo gps plugin can now read the world's datum making models using the plugin more portable.
https://github.com/tu-darmstadt-ros-pkg/hector_gazebo/commit/8e2a2a4664330387b68d31f42d0c4de290d33e80

### DIFF
--- a/vrx_gazebo/worlds/sandisland.xacro
+++ b/vrx_gazebo/worlds/sandisland.xacro
@@ -15,8 +15,8 @@
     <spherical_coordinates>
       <surface_model>EARTH_WGS84</surface_model>
       <world_frame_orientation>ENU</world_frame_orientation>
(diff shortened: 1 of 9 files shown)


bsb808:
@rolker - Apologies for letting this languish.   Now that we are moving the simulation to the new Syndney venue, I think changing the GPS lat/lon values is moot.  

We are incorporating your suggestion to make use of the new capability for the Hector GPS plugin to make use of the Gazebo spherical_coordinates parameters (PR #354), which will be much less prone to mistakes!

I'm closing for now, but feel free to re-open if I'm missing something.

PR #247, opened 2021-07-06
Migrate to Noetic and Gazebo11

crvogt:
#### Mesh Updates
Empty fields in `*.dae` have been pruned. In Gazebo11, signifying an empty field in a following forward slash format, ie. `<librarynotes/>`, would result in an error with RViz.

#### tf_prefix
The global `namespace` variable is used in favor of the deprecated `tf_prefix`, apart from usage as a parameter in the `robot_localization` node.

#### RViz Examples
RViz examples are now confirmed to work with new config files in both Gazebo9 and 11

### To Test
Run `roslaunch vrx_gazebo vrx.launch`, in another terminal tab `roslaunch wamv_gazebo localization_example.launch`, and finally `roslaunch wamv_gazebo rviz_vrx.launch` in another tab.

The resulting RViz window should look like the following:
Screenshot from 2021-07-08 15-47-23


caguero (commented):
~~It looks good so far! Could you also update the `docker/Dockerfile` to use Ubuntu 20.04, Gazebo 11 and ROS Noetic?~~ Never mind, this seems to be ready already.

caguero (approved):
I confirm that it worked for me in Noetic inside Docker!

PR #248, opened 2021-07-06
Noetic docker

M1chaelM:
To test, please work through the updated VRX [Docker Install Tutorial](https://github.com/osrf/vrx/wiki/docker_install_tutorial).


bsb808:
It works for me.  The run and join scripts are much simpler now!

I did make some modifications to `run.bash` in 9aae4469360217a94772692a65487fd11671b339 to make use of the CLI argument and parameterize the container name based on the image name - instead of hardcoding.

bsb808 (approved):
See comment in conversation.  Good to go on my end.

PR #253, opened 2021-07-08
Sydney Regatta Centre world

caguero:
vrx_22

Main features:

* New world based on the Sydney Regatta Centre.
* Updated lat/lon GPS reference.
* I've removed from this world the 2019 elements (buoys, docks, etc.) for not confusing teams. I'll add them back when we have the 2022 versions.

A few notes:

* As you can see there's an issue with the trees. The transparency of the leaves is not working and you see the evident four planes of the trees. I know the reason but it'll require a few days for me to fix it as we need to tweak the main `.dae` mesh.

* Typically, the main launch file is `vrx.launch`. I renamed the old `vrx.launch` file to be `vrx_2019.launch` from now on. `vrx.launch` will be the `2022` version.

For testing it:
```
```


M1chaelM:
Thanks @caguero ! The arena launches for me and looks really great. I think we should fix the trees before we merge though. After merging, we'll need to update the current tutorials to replace vrx.launch with vrx_2019.launch so they still work as expected. This should be easy--just wanted to leave myself a note here to do it.

caguero:
> Thanks @caguero ! The arena launches for me and looks really great. I think we should fix the trees before we merge though. After merging, we'll need to update the current tutorials to replace vrx.launch with vrx_2019.launch so they still work as expected. This should be easy--just wanted to leave myself a note here to do it.

It should be fixed in 2125622!

vrx_22_trees

M1chaelM (approved):
Nice! This is working for me on the host and I also see the tree issue is resolved, so I think it's good to go. Running in the Docker image still produces a crash because it can't find files that are included within the wamv_gazebo macro, so I'm still trying to get to the bottom of that but I think it's a separate issue.

Issue #250, opened 2021-07-07
`vrx.launch` fails in our docker noetic container


M1chaelM:
Attempting to launch VRX with:
```
roslaunch vrx_gazebo vrx.launch
```
fails with the following error:
```
unknown macro name: xacro:wamv_camera
when processing file: /home/username/vrx_ws/src/vrx/wamv_gazebo/urdf/wamv_gazebo.urdf.xacro
RLException: Invalid <param> tag: Cannot load command parameter [wamv/robot_description]: command [['/opt/ros/noetic/lib/xacro/xacro', '--inorder', '/home/username/vrx_ws/src/vrx/wamv_gazebo/urdf/wamv_gazebo.urdf.xacro', 'locked:=false', 'thruster_config:=T', 'vrx_sensors_enabled:=true', 'namespace:=wamv']] returned with code [2]. 

Param xml is <param name="$(arg namespace)/robot_description" command="$(find xacro)/xacro --inorder '$(arg urdf)'          locked:=$(arg wamv_locked)               thruster_config:=$(arg thrust_config)          vrx_sensors_enabled:=$(arg vrx_sensors_enabled)          namespace:=$(arg namespace) "/>
The traceback for the exception was written to the log file
```
This same command succeeds in launching gazebo on the host using the host-based install instructions, even without rebuilding vrx. Maybe there is a missing dependency in the container?

M1chaelM:
Continuing to document this issue:
Running the xacro command directly fails in the docker image whenever the argument `vrx_sensors_enabled:=true` is included. This argument indirectly sets `camera_enabled` to true, which triggers the error. A simpler test is:
```
/opt/ros/noetic/lib/xacro/xacro /home/username/vrx_ws/src/vrx/wamv_gazebo/urdf/wamv_gazebo.urdf.xacro camera_enabled:=true
``` 
This works fine on the host, and fails in the Docker container with a similar error:
```
unknown macro name: xacro:wamv_camera
when processing file: /home/username/vrx_ws/src/vrx/wamv_gazebo/urdf/wamv_gazebo.urdf.xacro
```

M1chaelM:
* All the files in the `sensors` directory are included by `macros.xacro`.
* `macros.xacro` is included by `wamv_gazebo.xacro` at line 6
* `wamv_gazebo.xacro` is included in `wamv_gazebo.urdf.xacro` at line 36.

M1chaelM:
[...]

PR #255, opened 2021-07-09
Cleaning previous to 1.4 release

caguero:
* I don't think we need the `docker/keys` directory anymore if we're downloading the keys from the key server. This was a bit unreliable for a while but it's cleaner if we don't see errors anymore. 

* I also added @M1chaelM's suggestion for fixing the xacro issue inside Docker.

* I removed a warning in `usv_gazebo_plugins/src/shape_volume.cc`.


M1chaelM (approved):
Thanks for cleaning up the Dockerfile. These changes look good and I'm now able to run `vrx.launch` on both host and Docker image, so I think this also resolves issue #250.

PR #256, opened 2021-07-09
adding latest noetic image to ci tests

M1chaelM:
This adds tests against the new software requirements: Focal/noetic/gazebo11. I didn't remove the previous tests, but we may eventually want to do this if they get cumbersome (we are still going back to Gazebo 7).


M1chaelM:
It looks like the new test is currently failing because it is implementing better static checks. To get this to pass we should fix the code.

caguero:
> It looks like the new test is currently failing in some cases because it is implementing better static checks. These are finding some things we should fix. It also looks like cppcheck needs to be configured so it recognizes Gazebo macros.

I tried to fix the style issues in 2f8cf36 but there's still something we need to tweak.

caguero (approved):

PR #263, opened 2021-07-12
[noetic] Fix joint reference

kledom:
This fixes the problem that no pointclouds are being published.


crvogt (approved):
Looks good, will add more. Thanks

PR #264, opened 2021-07-14
2022 buoys

caguero:
Here's the set of buoys used for RobotX 2022:
buoys_22

Here are the collisions of the two buoy types:
buoys_collisions

The buoy colors are based on the specification from the Marine Buoys Australia website. See [here](https://marinebuoysaustralia.com/wp-content/uploads/2020/12/Buoy-colours-scaled-1.jpg).


M1chaelM (approved):
Confirming I was able to load all 6 new buoys, and the collisions match what you describe in the PR. I think this is good to go!

Issue #261, opened 2021-07-09
Update to Python files to use Python3


M1chaelM:
## Context
* Since we are moving to Ubuntu 20.04, python 2 is no longer installed by default.
* We could install it, but this adds dependencies and some amount of confusion keeping the versions straight.
* We only have about a dozen python files and it is not too hard to move them to python 3 (in many cases this is just a matter of changing outdated executable paths at the top of the script).

## Proposed
Let's make our code work with Python 3.

crvogt:
@M1chaelM This seems like something I can take care of.

M1chaelM:
It looks like our acoustic pinger tutorial may rely on this changes, since the `pinger.launch` file fails with the Python error:
```
ImportError: No module named yaml
```
This comes from both the `pinger_visualisation.py` and `set_pinger_position.py` nodes. The yaml module is definitely installed so I think these are trying to run python 2.

crvogt:
@M1chaelM I found that by simply modifying the shebangs at the top of the python files to python3 the pinger tutorial work fine. I haven't tested in Melodic though. I think ideally there'd be a parameter to use one or the other, but for a quick fix, this does work. Thoughts?

M1chaelM:
I think changing to force running in Python 3 should be fine. This will just mean that people running in Melodic need to install Python 3, which shouldn't be too much of a burden. It looks like @caguero did find a couple places where the code needs to be updated to eliminate things that aren't supported in Python 3, so it may be worth skimming through to look for these in case they don't pop up right away in testing.

crvogt:
@M1chaelM Sounds good!

M1chaelM:
Fixed in PR #265 and #266.

PR #265, opened 2021-07-15
Python updates

caguero:
While going though the tutorials I noticed that some Python files needed to be updated.


M1chaelM:
@caguero Thanks! I think this partially addresses issue #261. I will review it now.
FYI, @crvogt this is pretty minor but might affect changes you are making.

M1chaelM (approved):
Looks good!

vrx_gazebo/src/vrx_gazebo/compliance.py
```diff
@@ -183,7 +183,7 @@ def find_boxes(box_yaml):
```
M1chaelM, line 183: In lines 157 and 172-177 above, are we intending to use integer division? It might not matter, but I ust want to make sure...

M1chaelM (approved):
Assuming these latest changes look good I think this should be ready to merge.

vrx_gazebo/src/vrx_gazebo/generate_worlds.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
caguero, line 1: Are these changes backwards compatible? If so, I think `python` in the shebang will make it compatible with Python2.

crvogt: @caguero Verifying that now...

M1chaelM: In the default setup of Focal the following give an error:
```
$ python
bash: python: command not found
$ /usr/bin/env python
env: ‘python’: No such file or directory
$ /usr/bin/python
env: ‘python’: No such file or directory
```
In other words, there is no longer a versionless "python" alias that points to anything. You have to either set up the alias or explicitly ask for a version of python that's installed on your system. Since we have updated requirements to Focal, 
python3 is installed by default, so I agree with updating to point to this.

Users still working on Bionic should have no problem as long as they have Python 3 installed. If someone is really trying to avoid installing Python 3 and use Python 2 only, they would have to do some work, but in my opinion we should wait to address this until we understand the use case. For most people it seems pretty reasonable to just add Python 3 as a dependency.

M1chaelM: @caguero OK to merge this?

caguero: Thanks for the verification, it looks good to me as it is then.

PR #266, opened 2021-07-15
Modified file shebangs to python3

crvogt:
Specific example now working: `roslaunch vrx_gazebo vrx.launch` -> `roslaunch vrx_gazebo pinger.launch` -> `roslaunch wamv_gazebo localization_example.launch` -> `roslaunch wamv_gazebo rviz_vrx.launch`

### DIFF
--- a/vrx_gazebo/nodes/key2thrust_angle.py
+++ b/vrx_gazebo/nodes/key2thrust_angle.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import rospy
 from std_msgs.msg import Float32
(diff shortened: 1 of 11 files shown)


M1chaelM (approved):
I can verify that this solves several problems. Since we are not requiring/expecting python 2 I don't see how it could create new problems. I also ran a few sanity checks:
* Verified this resolves the error in `set_pinger_position.py` and `pinger_visualization` called by `pinger.launch`, so the acoustic pinger tutorial now looks good.  
* `usv_keydrive` and `usv_joydrive` also still seem to be running fine. 
* `generate_avoid_obstacles_buoys` is also producing the expected output.
In addition to the above runtime tests, I ran `pylint --py3k` on all files listed. All look good except `compliance.py`, `utils.py` and `generate_worlds.py`. These are being fixed in PR #265.

So overall I think we are ready to merge this. If we come across additional Python 3 errors in the future we can revisit this.

PR #267, opened 2021-07-16
Adjusted grid location in RViz to correspond with wamv location

crvogt:


wamv_gazebo/config/rviz_vrx.rviz
```diff
@@ -47,9 +43,9 @@ Visualization Manager:
       Name: Grid
       Normal Cell Count: 0
       Offset:
-        X: 0
-        Y: 0
-        Z: 0
+        X: 532
+        Y: -162
+        Z: -0.11999999731779099
```
M1chaelM, line 48: Instead of changing the offset, can we set the reference frame to wamv/baselink? Then it should center the grid around the vehicle instead of at the origin of the world.

crvogt: @M1chaelM fixed

M1chaelM (approved):
OK, this is working for me. Digging in to this a little, the original version looks like it used wamv/odom for its reference frame. This link is set at the origin of the world in `/urdf/sensors/wamv_p3d.xacro`. Because the origin is far away from the default position of the WAMV in this world, the demo ends up looking weird--not a big technical problem but it could be confusing. We're fixing this by setting the base link to wamv/base_link, which should be fine. Next, we need to update the tutorial--especially the part at the bottom that says the default link is "odom" since this is no longer true. The explanation of how to change the reference frame is still useful though.

PR #270, opened 2021-07-16
update changelog for v 1.4

M1chaelM:
Just keeping the Changelog up to date.


M1chaelM:
@caguero @crvogt Apologies, but it seems I no longer have the power to make a change like this without a review. Please approve when you can!

caguero (approved):

PR #304, opened 2021-08-04
Remove docker directory now that we have dockwater

bsb808:
I believe we also need to update the wiki https://github.com/osrf/vrx/wiki/docker_build_run_tutorial

@M1chaelM I think you would have a better sense of how to keep this consistent across the dave, vrx, dockwater set of projects.


M1chaelM (approved):
Looks good. The wiki is up to date.

PR #305, opened 2021-08-04
Add utility script for rapidly shutting down gazebo

bsb808:
This is a hack I use to be quickly shutdown the sim...

```
rosrun vrx_gazebo killgazebo.sh
```

Is there a better way?


caguero:
Isn't the usual `CTRL-C` working for you? 90% of the time works for me and when it doesn't I manually run the `kill` command for the server.

caguero (commented):
Not sure if it's strictly necessary but maybe we need this line in the `CMakeLists.txt`:

```
install(PROGRAMS scripts/kill_gazebo.sh
  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
```

You could probably move it to the `scripts` directory as well.

bsb808:
The script is just a convenience.  It shuts down gazebo (client and server) much faster than the ctrl-c interrupt.

bsb808:
> Not sure if it's strictly necessary but maybe we need this line in the `CMakeLists.txt`:
> 
> ```
> install(PROGRAMS scripts/kill_gazebo.sh
>   DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
> ```
> 
> You could probably move it to the `scripts` directory as well.

ROS seems to find the executable in any subdirectly of the package.  For me, this works....
rosrun vrx_gazebo killgazebo.sh

caguero (approved):

Issue #280, opened 2021-07-22
Implement the deliver aspect of the "Dock and deliver" task


caguero:
* Model a basic launcher (it could be as simple as a cylinder. We can potentially have Cole to model something that looks nicer).
* Write a plugin that let the teams to throw a projectile.
  * Define the API: It could go from exposing some velocity control at the end of the launcher and optionally to add pan/tilt capabilities. We should ask Arjo about the common API in RobotX. Not sure if it's common to control pan/tilt in this actuator. 
  * Implement the functionality in the plugin

M1chaelM:
This issue was resolved with #311 and #316.

PR #311, opened 2021-08-10
New ball shooter plugin for the "Scan and Dock and Deliver" task

caguero:
See issue #280.

How to test it:

Terminal 1:
```
```

Terminal 2:
```
rostopic pub /wamv/shooters/ball_shooter/fire std_msgs/Empty "{}" --once
```

You should observe how the blue ball is fired.
ball_shooter

Feel free to play with the plugin parameters in `ball_shooter.xacro`  for testing.

Note about the ball shooter model: For now, the model is just a box. This will be replaced in the future when a nicer model is ready. It shouldn't be any functional difference in any case.


M1chaelM:
The gazebo 7 test is failing. Since this is very old now, I've changed the merge requirements to make it optional and added the gazebo 11 test to the list of required checks. I'm also updating the test images just to make sure they are correct.

vrx_gazebo/include/vrx_gazebo/ball_shooter_plugin.hh
```diff
@@ -0,0 +1,114 @@
...
+///                               firing the shooter. Required parameter.
+///                 * <frame> If present, the <pose> parameter will be in the
+///                           frame of this link/model. Otherwise the world
+///                           frame is used. Optional parameter.
+///                 * <pose> - Pose of the projectile right before being shot.
+///                            Optional parameter. Default to {0 0 0 0 0 0}.
+///
+/// * Optional parameters:
+/// <num_shots> - Number of shots allowed. Default to UINT_MAX.
+/// <shot_force> - Force (N) applied to the projectile. Default to 250 N.
```
M1chaelM, line 51: What do you think about including this tag explicitly in the example and also the .xacro file we are providing ? Out of all the parameters, it seems like the one that users are most likely to want to adjust. Even if it is redundantly set to 250 it might make it easier to quickly pick up how to use this plugin.

wamv_gazebo/urdf/ball_shooter/ball_shooter.xacro
```diff
@@ -0,0 +1,41 @@
...
+    </joint>
+    <gazebo>
+      <plugin name="ball_shooter_plugin" filename="libball_shooter_plugin.so">
+        <projectile>
+          <model_name>blue_projectile</model_name>
+          <link_name>link</link_name>
+          <frame>wamv/ball_shooter_link</frame>
+          <pose>0.2 0 0 0 0 0</pose>
+        </projectile>
+        <topic>${namespace}/${shooter_namespace}${name}/fire</topic>
```
M1chaelM, line 37: This is where I was thinking we could change to something like:
```
       <shot_force>250</shot_force>
       <topic>${namespace}/${shooter_namespace}${name}/fire</topic>
```

M1chaelM (approved):
I think this looks great! Thanks for putting it together. I added one suggestion about putting the shot_force explicitly in the ball_shooter.xacro, if you agree that could make things easier for users.

A couple other notes:
* I tested varying the shot force and number of shots. These seemed to be working as expected.
* I see what you mean about re-using the same ball. I think this is fine. I guess it's possible that teams will want to use a strategy of shooting a bunch of shots in rapid succession if they are running out of time, and this might be OK to allow if we decide to limit the number of shots (say, to 3 or whatever would reasonably fit). I think we can adjust this later if needed though.

M1chaelM:
One more note: after updating the test images this is still not building on Gazebo 7. I'm not sure whether we should try to fix this or not. Thoughts?

vrx_gazebo/include/vrx_gazebo/ball_shooter_plugin.hh
caguero: Good idea! Added in 07609f6.

caguero:
> One more note: after updating the test images this is still not building on Gazebo 7. I'm not sure whether we should try to fix this or not. Thoughts?

I needed to add a few `ifdefs` to the code. Not very elegant but it should work. Waiting for CI...

PR #313, opened 2021-08-13
Update Station Keeping Task

crvogt:
##### Updates the station keeping task:
- Set in sydneyregatta world
- GPS headings updated
- Goal position updated

To run:
`roslaunch vrx_gazebo station_keeping.launch`

To move WAM-V via keyboard:
`roslaunch vrx_gazebo usv_keydrive.launch`

To monitor score:
`rostopic echo /vrx/station_keeping/pose_error`

##### Task Summary

Navigate to the goal pose and hold station. The best solutions will minimize the difference between the goal pose and the actual pose of the vehicle over the duration of the task.

Start the example: roslaunch vrx_gazebo station_keeping.launch
Subscribe to the task-specific topics provided by the stationkeeping scoring plugin:
The station-keeping goal (given as a geographic_msgs/GeoPoseStamped message):
rostopic echo /vrx/station_keeping/goal
The current position error values:
rostopic echo /vrx/station_keeping/pose_error
rostopic echo /vrx/station_keeping/rms_error
For implementation details, see "4.1.1. Task 1: Station-Keeping" in the Competition and Task Descriptions, or refer to the stationkeeping scoring plugin.

### DIFF
--- a/vrx_gazebo/launch/station_keeping.launch
+++ b/vrx_gazebo/launch/station_keeping.launch
@@ -13,12 +13,12 @@
   <arg name="namespace" default="wamv"/>
 
   <!-- Initial USV location and attitude-->
-  <arg name="x" default="158" />
-  <arg name="y" default="108" />
+  <arg name="x" default="532" />
+  <arg name="y" default="-162" />
   <arg name="z" default="0.1" />
   <arg name="P" default="0" />
   <arg name="R" default="0" />
-  <arg name="Y" default="-2.76" />
+  <arg name="Y" default="-2.15" />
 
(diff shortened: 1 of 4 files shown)


j-herman (approved):
Verified the scenario starts up correctly.  GPS appears to be working correctly, error calculations look good, and all topics are being published.
Not a big issue and not required for merging, but we might want to revisit the topic name for the mean error.  Currently published at "/vrx/station_keeping/rms_error".  This suggests to me that we're calculating the instantaneous root mean square error, when what we're really publishing/scoring is the long-term average error.

PR #314, opened 2021-08-13
Update Wayfinding Task

crvogt:


j-herman (approved):
All looks good!  Verified positions of the WAM-V and waypoints.  Message topics all publish and update correctly.

Issue #291, opened 2021-07-22
Create a launcher model to be added to the WAM-V


caguero:
This is the launcher to be used in the "deliver" part of the "Dock and Deliver" task.

caguero:
Initial look of the model:

![ball_shooter](https://user-images.githubusercontent.com/1440739/129270734-0be1f837-8055-41ac-9383-22370aca624a.jpg)

PR #316, opened 2021-08-13
Add ball shooter model.

caguero:
See issue #291.

This pull request adds a stand alone ball launcher model. I also updated `ball_shooter.xacro` to use the visuals of this model.

Here's a picture of the standalone model showing its inertia.

shooter_inertia

Note that the origin of the model is at the base with `X` pointing forward. However you'll notice that the cylinder has a -15 deg. pitch. I added a few offsets to compensate for that. That way, when we specify the pose and orientation where we want the ball shooter in the WAM-V, the pitch angle that we specify will match the angle of the cylinder (see the `-0.261799` value below). 

```
<!-- Add ball shooter (default pitch angle: -15 deg) -->
      <xacro:wamv_ball_shooter name="ball_shooter" position="0.54 0.30 1.296" orientation="0 -0.261799 0"/>
```

There's also a helper visual that I use to figure out the `offset` parameter for the `BallShooter` plugin. This visual is there to show the pose where the ball is teleported before being shot. Here's how it looks.

shooter_ball


M1chaelM:
@caguero I'm seeing two problems so far. The inertia for the model doesn't match screenshot:
Screenshot from 2021-08-17 16-21-58
Maybe it's perpendicular. Also, when I tried to add a standalone model using the Insert tab, nothing happened. I'm able to add other models; just not this one.

The first problem seems like an easy fix as long as you are seeing the same thing. I'm not sure what would cause the second one.

caguero:
> @caguero I'm seeing two problems so far. The inertia for the model doesn't match screenshot:
> Screenshot from 2021-08-17 16-21-58
> Maybe it's perpendicular. Also, when I tried to add a standalone model using the Insert tab, nothing happened. I'm able to add other models; just not this one.
> 
> The first problem seems like an easy fix as long as you are seeing the same thing. I'm not sure what would cause the second one.

Good catch. The first issue should be fixed in 15c8f34.

shooter_inertia

The second issue is related with submeshes. I'm still looking into it.

caguero:
> > @caguero I'm seeing two problems so far. The inertia for the model doesn't match screenshot:
> > Screenshot from 2021-08-17 16-21-58
> > Maybe it's perpendicular. Also, when I tried to add a standalone model using the Insert tab, nothing happened. I'm able to add other models; just not this one.
> > The first problem seems like an easy fix as long as you are seeing the same thing. I'm not sure what would cause the second one.
> 
> Good catch. The first issue should be fixed in [15c8f34](https://github.com/osrf/vrx/commit/15c8f34c9bf5ffee8bc16575ccc0034378a10c8a).
> 
> shooter_inertia
> 
> The second issue is related with submeshes. I'm still looking into it. Here's a better error:
> 
> ```
> [Err] [Visual.cc:374] Ogre Error:OGRE EXCEPTION(6:FileNotFoundException): Cannot locate resource /home/caguero/vrx2022_ws/src/vrx/vrx_gazebo/models/ball_shooter/meshes/ball_shooter.dae::Launcher in resource group Autodetect or any other group. in ResourceGroupManager::openResource at /build/ogre-1.9-kiU5_5/ogre-1.9-1.9.0+dfsg1/OgreMain/src/OgreResourceGroupManager.cpp (line 753)
> [Err] [Visual.cc:375] Unable to create a mesh from /home/caguero/vrx2022_ws/src/vrx/vrx_gazebo/models/ball_shooter/meshes/ball_shooter.dae
> ```

The second issue should be fixed now. See a1478f4.

PR #321, opened 2021-08-24
Add animal buoys.

caguero:
Adding the new models to the animal buoys.

crocodrile

platypus

turtle

I also converted the collision elements to simple shapes:

crocodrile_collision

platypus_collision

turtlebot_collision


M1chaelM:
These look great! I will review today.

Issue #322, opened 2021-08-24
Revise Pose Error distance for Station keeping, Wayfinding and Gymkhana tasks


M1chaelM:
We are adopting a new method of calculating combined pose error distance:
```
k = 0.75
d = sqrt((x0-x_i)^2 + (y_0-y_i)^2)
alpha = abs(h_0 - h_i) // raw angle difference: [0,2pi]
h = PI - abs(alpha-PI)) // shortest heading distance: [0,pi]
E = d + (k^d)* h
```
Where:
* x0,y0 are the cartesian coordinates of the goal
* xi,yi are the cartesian coordinates of the vehicle at time step i
* h_0 is the yaw of the goal in radians
* h_i is the yaw of the USV at time i
* k^d is a weighting term that grows inversely proportional to distance. A value of k = 0.75 causes the header distance to begin to have an impact at a range of about 10 meters.
We have updated the documentation to describe this. Next we need to update the scoring plugins for the three tasks. In the Gymkhana task the heading score needs to be discarded completely, but this is documented in a separate issue (#317).

Note to self that the range of the yaw angle is determined with the output of C++'s atan2 function, which is [-pi,pi].

Also needed:
* Change ROS topics for station keeping and gymkhana to match updated API in [Task Descriptions](https://docs.google.com/document/d/1a2-L0UeZKFq4RX283a22PO2gGTAXVIqV/edit).

PR #325, opened 2021-08-26
Fixing pose error metric to match documentation.

M1chaelM:
Addresses issue #322 .

To review:
* Double check the scoring math in modified lines of `station_keeping_scoring_plugin.cc` and `wayfinding_scoring_plugin.cc` to make sure it matches the [Task Descriptions document](https://docs.google.com/document/d/1CQTHCFdMgHdQASS7kwGIER1zMarIBydOXv5wL9lEpR0/edit).
* Run the [task tutorials](https://github.com/osrf/vrx/wiki/tutorials) for station_keeping, wayfinding and gymkhana and check that they are behaving as described in the documentation (especially that the topics are correct).

Note that there is still a problem with the gymkhana scoring, which is that it is orientation-sensitive but it should not be.  This will be addressed in a separate PR.


j-herman:
Math appears to be correct per description in issue #322 but I will wait to close out the review until I can validate against the Task Description document (need access to google drive).

Ran all three task tutorials.  General behavior appears to be correct, and scoring appears qualitatively consistent with description for station-keeping and wayfinding.  I found a few issues with topic names that most likely need to be corrected in the tutorials rather than the code.

Topic needs to be updated in stationkeeping task description.  Step 3 calls out:
/vrx/station_keeping/rms_error
New code is publishing on /vrx/station_keeping/mean_pose_error

I was unable to trigger anything on the two gymkhana debug/contact topics (assuming those are intended to indicate collisions?) - recommend describing better in the wiki what these are supposed to tell you, or removing the references in the wiki if these are not in use.
/vrx/gymkhana_blackbox/debug/contact
/vrx/gymkhana_channel/debug/contact

Typo in the gymkhana task:
"/vrx/gymkhana_blackbox/pose_error provides the root mean square error."
The current correct topic appears to be /vrx/gymkhana_blackbos/rms_error but I'm not sure if that will also be changing to mean_pose_error or a different name.

M1chaelM:
@j-herman Please try the new link to the Task Descriptions document. It changed this morning and I forgot to update the issue.

M1chaelM:
@j-herman Thanks for taking a look at this. Once you have access to the Task Descriptions document, please use that as the authoritative explanation. If the wiki disagrees with the documentation, the wiki probably needs to be updated (thanks for catching this!).

I went through quickly and corrected the problems you pointed out in the Wiki. I'm sure they could use a second look though. Thanks again!

M1chaelM:
@j-herman Thanks for catching it. I want to fix it in this PR. Should be fixed with the last commit (I also changed the element name to avoid confusion.)

j-herman (approved):

Issue #276, opened 2021-07-22
Create task 4 - Wildlife Encounter and Avoid


caguero:
This is a new task.
* Load Sydney Regatta.
* Update the starting pose of the WAM-V
* Need to coordinate with Open Robotics artist to create models of wildlife.
* Generate some courses with wildlife.
* Decide on scoring
* Create scoring plugin
* Create a tutorial here: https://github.com/osrf/vrx/wiki/vrx_2022-wildlife_task
* Review [Task Descriptions document](https://docs.google.com/document/d/1a2-L0UeZKFq4RX283a22PO2gGTAXVIqV/edit) to verify the API and description are correct.

PR #326, opened 2021-08-27
Wildlife scoring plugin - part 1

caguero:
See issue #276.

This scoring plugin isn't trivial, so I decided to split it in two pull request. This is the first one:

* It adds a new world (`wildlife_task.world.xacro`).
* It adds a new launch file for launching the previous world (`wildlife.launch`).
* It adds a new model (`animal_buoys`) that combine all the animal buoys into a single model. The reason for this is that it will be easier in the future to create different combinations for each run. Every configuration will be captured into a separate model containing also the `FollowPlugin` parameters. For now, I didn't move the plugins to this new model.
* It adds a new scoring plugin (`wildlife_scoring_plugin`).

The functionality added so far is:
  * The plugin parses all the SDF parameters. 
  * The plugin publishes all the animal poses as documented in the rules.
  * The plugin instantiates a new `Buoy` object for each buoy (animal) and updates its internal state. There are four states: NEVER_ENGAGED, NOT_ENGAGED, ENGAGED and CIRCUMNAVIGATED.
 
The part 2 will include:
   * The logic that decides when a buoy has been circumnavigated.

For testing:
```
roslaunch vrx_gazebo wildlife.launch extra_gazebo_args:=--verbose
```

```
```

You should see messages of this type:
```
[Dbg] [wildlife_scoring_plugin.cc:63] [WildlifeScoringPlugin::Buoy] crocodile_buoy::link Transition from NOT_ENGAGED to ENGAGED
```

For testing the buoy poses you can type:
```
rostopic echo /vrx/wildlife/animals/poses
```

### DIFF
--- a/vrx_gazebo/CMakeLists.txt
+++ b/vrx_gazebo/CMakeLists.txt
@@ -183,6 +183,18 @@ install(TARGETS navigation_scoring_plugin
   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
 
+# Plugin for scoring the wildlife challenge task.
+add_library(wildlife_scoring_plugin src/wildlife_scoring_plugin.cc)
+target_link_libraries(wildlife_scoring_plugin
+  ${catkin_LIBRARIES}
+  scoring_plugin)
+add_dependencies(wildlife_scoring_plugin ${catkin_EXPORTED_TARGETS})
+install(TARGETS wildlife_scoring_plugin
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
+
 # Plugin for scoring the station keeping challenge task.
 add_library(stationkeeping_scoring_plugin src/stationkeeping_scoring_plugin.cc)
 target_link_libraries(stationkeeping_scoring_plugin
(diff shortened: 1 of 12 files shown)


M1chaelM:
@caguero Did you mean to assign me as a reviewer?

caguero:
> @caguero Did you mean to assign me as a reviewer?

Yes, I probably missclicked.

M1chaelM:
@caguero It looks like `wildlife_task.launch` might not have been added to your commit. I don't see this file.

For consistency with the other task names, could you also rename it to `wildlife.launch` when you add it?

M1chaelM (approved):
OK, this seems to be working for me now. I'm seeing correct debug messages for transitions from NEVER ENGAGED to ENGAGED on the first encounter with the animal, and from ENGAGED to NOT ENGAGED and the reverse when moving in and out of range.

## Notes:
* The task proceeds as expected from the OnReady to OnFinished states, but then crashes with a segfault. It seems like Gazebo often terminates with a segfault crash so this may be normal; just want to check.
* I can see updates of the lat long positions of the animals. This seems to be working but I don't have a good way to verify that the coordinates are correct so I didn't check them.
* Similarly, although the transitions in animal state seem to be working I didn't check that they are happening at exactly the right distances. I did read through the conditionals in the scoring plugin and they look right to me.
* The crocodile happens to be spawned too close to the wamv to avoid it, so the transition to ENGAGED occurs before the vehicle can move. I'm not sure if we want to wait until OnReady to start checking for these transitions, or just make sure we put the animals farther away when we design the course.

caguero:
> ## Notes:
> * The task proceeds as expected from the OnReady to OnFinished states, but then crashes with a segfault. It seems like Gazebo often terminates with a segfault crash so this may be normal; just want to check.

This is expected.

> * I can see updates of the lat long positions of the animals. This seems to be working but I don't have a good way to verify that the coordinates are correct so I didn't check them.

My check was that the values were close enough to the initial lat/long of the WAM-V and they seem close enough.

> * Similarly, although the transitions in animal state seem to be working I didn't check that they are happening at exactly the right distances. I did read through the conditionals in the scoring plugin and they look right to me.

👍

> * The crocodile happens to be spawned too close to the wamv to avoid it, so the transition to ENGAGED occurs before the vehicle can move. I'm not sure if we want to wait until OnReady to start checking for these transitions, or just make sure we put the animals farther away when we design the course.

Agreed, the animal locations and trajectories are not final. I'll move them to reasonable  locations in the 2nd part.

PR #328, opened 2021-09-03
replace broken utility function

M1chaelM:
Our configure_wamv.py was using a custom string-replace function to create file names with .xacro extensions based on the names of .yaml files. This function only works if there are no other "." characters anywhere in the path. I've replaced it with a standard pattern using os.path.splitext. Since we're already importing os anyway, I can't think of any reason not to do this.

A side note: debugging this took me on a little tour through the portion of our code that generates xacros from yaml. The tour was a little more exciting than it probably should be. In a future release it would be good to take another look at this.


caguero (approved):
It looks good to me!

Issue #281, opened 2021-07-22
Verify scoring plugins of tasks 1, 2, 3 and 5


caguero:
Test these tasks with the scoring plugins running and verify that the scores match what we describe in the 2022 "VRX Competition and Task Description" document.

M1chaelM:
This task depends on the prior completion of the following tasks:
* #272
* #273
* #274
* #277

M1chaelM:
Slightly redundant with #293. The intent of this issue was just to check that scoring is working for tasks where the scoring plugin had already been developed (the omitted tasks are the ones that we made new scoring plugins for this time).

j-herman:
I am working on task 1 for both the scoring plugin and tutorial verification

Issue #293, opened 2021-07-22
Test all the tutorials


caguero:
Use this ticket to organize who is testing each tutorial.

M1chaelM:
Following [this testing plan](https://docs.google.com/spreadsheets/d/1s60ptIg2Dau96Y_Z0M9AvalAJawofmDaGbQI9WT1QZY/edit#gid=0).

j-herman:
I am working on task 1 for both the scoring plugin and tutorial verification

PR #330, opened 2021-09-03
Wildlife scoring plugin - part 2

caguero:
See issue #276.

This pull request is the final part of pull request #326. All the scoring aspects of the task should be covered here.

Here are some relevant notes:
* I removed the `animal_model` because for some reason the `follow_plugin` doesn't work well with this approach. It was faster for me to go back to the original approach using three models than looking into the issue.
* The task finishes when you consume all the running time (timed out) or when you circumnavigate all of the circumnavigable buoys.
* I fixed the obstacle detection in this pull request. In this task, when an obstacle is hit with the WAM-V, the buoys not fully circumnavigated are invalidated and you need to start over with these buoys (it should be just one in practice).
* The time bonus is applied at the end of the task.
* I moved the animals a bit far away to avoid overlaps.
* The task 4 wiki page has been updated.

For testing:
```
roslaunch vrx_gazebo wildlife.launch extra_gazebo_args:=--verbose
```

```
```

Try circumnavigating the buoys. You should see messages of this type:
```
[Dbg] [wildlife_scoring_plugin.cc:191] [WildlifeScoringPlugin::Buoy] turtle_buoy::link Virtual gate crossed counterclockwise! (25% completed)
```

For checking the score:
```
rostopic echo /vrx/task/info
```


M1chaelM (approved):
Circumnavigation seems to be working well and the messages are very handy! Scoring also looks good--it tracks the run time and subtracts the bonus at the end (90s max).

I didn't extensively test different scoring scenarios but I think this is OK because we are planning to double check them this week to resolve issue #281 (and also #293).

Issue #317, opened 2021-08-16
Gymkhana scores are incorrect


M1chaelM:
## Heading error problem
The Gymkhana scoring plugin relies on the stationkeeping scoring plugin to calculate the RMS pose error between the USV and the blackbox over the duration of the run. However, the stationkeeping scoring plugin was intended to calculate a combined position and heading error, and the blackbox does not have a heading. 

This could work correctly if the heading error term were simply set to 0. What's happening instead is that the goal pose is being passed in with X, Y, Z coordinates, where Z is supposed to be a depth of -5 meters, but the stationkeeping scoring plugin is expecting this term to represent the goal yaw. The result is that the plugin interprets -5 as an angle and calculates the error distance between the actual heading of the USV and the angle of -5 radians throughout the entire task.

## Solution for heading error problem
One thing we could do that's consistent with our documentation of the stationkeeping scoring plugin is to modify the stationkeeping scoring plugin so it can take an argument that allows it to set the weighting function used for the heading error, and then pass in an argument that causes the weight to be set to 0. Right now the weighting function is simply the identity function and there is no way to modify it without recompiling.

## Obstacle penalty problem
We are using the navigation plugin to calculate the penalty for colliding with an obstacle. This plugin's default behavior is to add a time penalty for each obstacle. The penalty is added in real time and extends the total amount of time the team took to complete the navigation channel.

In the gymkhana task, the time penalty is irrelevant because the whole score is given in terms of average error distance from the black box, so this penalty is probably being ignored.

## Solution for obstacle penalty problem
We need to modify the gymkhana problem to track collisions over the entire task. At the end of the task, a 10 cm penalty should be added to the final run score for each collision.

M1chaelM:
Resolved with #337

PR #337, opened 2021-09-09
Fix gymkhana scoring issues

caguero:
See issue #317.

As described in the issue:

* I exposed the parameter `k` as an SDF parameter used in the station keeping scoring plugin. I named the parameter `heading_weight` but I'm not very convinced about it. Let me know if you can think of a better name.
* I added the logic to consider the WAM-V collisions to the gymkhana scoring plugin. We now have an extra SDF parameter to set the penalty for each collision. We now apply the penalties when the task finishes.

How to test it:

* The disable of the heading error is harder to test. The only quick way I can think of is to move closer to the black box, print the score continuously, and use the rotation tool to rotate the WAM-V verifying that there aren't big jumps in the score. Note that the scoring should change as the moving average keeps adjusting but we shouldn't see big jumps.

* For testing the obstacle penalty, you can drive the WAM-V through the channel,  then, hit one buoy,  print continuously the score and let the task finish. At the end you should see the correction. Here's an example:

```
---
name: "gymkhana"
state: "running"
ready_time: 
  secs: 10
  nsecs:         0
running_time: 
  secs: 20
  nsecs:         0
elapsed_time: 
  secs: 79
  nsecs:         0
remaining_time: 
  secs: 1
  nsecs:         0
timed_out: False
score: 93.83531551021375
---
name: "gymkhana"
state: "finished"
ready_time: 
  secs: 10
  nsecs:         0
running_time: 
  secs: 20
  nsecs:         0
elapsed_time: 
  secs: 80
  nsecs:         0
remaining_time: 
  secs: 0
  nsecs:         0
timed_out: True
score: 193.71494983741547
---
```

Note that for testing I modified the value of `running_state_duration` to `80` and `obstacle_penalty` to `100` to make it more evident.


vrx_gazebo/worlds/gymkhana.world.xacro
```diff
@@ -104,6 +105,8 @@
       <contact_debug_topic>/vrx/gymkhana_blackbox/debug/contact</contact_debug_topic>
       <!-- Keep Gazebo running after this sub-task is completed -->
       <per_plugin_exit_on_completion>false</per_plugin_exit_on_completion>
+      <!-- Disable heading error -->
+      <heading_weight>0</heading_weight>
```
M1chaelM, line 109: Consistent with other comments, can we change the name and make the argument for this true/false?

M1chaelM (changes requested):
See my inline comments about a small bug on line 197 of `stationkeeping_scoring_plugin.cc`.  Since we don't really need to be able to alter the value of `k` without recompiling, I think it might be simpler to introduce a new variable that is always either 1 or 0 depending on whether the heading error should be taken into account or not.  We could put this in the poseError equation, but I'm thinking it's probably more readable to just use a conditional.

To make sure I'm being clear I'm putting some inline comments about the specifics.

vrx_gazebo/worlds/gymkhana.world.xacro
caguero: Thanks for the suggestions. I changed to a boolean flag in 758c80c. I already started with a boolean flag approach, and then, I changed my mind to expose `k` :) . That's why I left unintentionally the parsing bug. It should be good to go!

M1chaelM (commented):
I tested that heading error appears to be turned off by teleporting to the goal location with various yaw values and checking the instantaneous error published to `/vrx/gymkhana_blackbox/pose_error`. For example:
```
gz model -m wamv -x 483.225 -y -295.584 -z 0 -Y -1.2
```
and
```
gz model -m wamv -x 483.225 -y -295.584 -z 0 -Y 0.2
```
both give extremely small errors. So it looks like the changes had the desire effect.

For the obstacle penalty I hit two buoys 13 times and ran out the score while sitting very close to the goal pose. My position relative to the goal caused the average score to go down by about 0.07 per time step. The last two messages in /vrx/task/info were:
```
name: "gymkhana"
state: "running"
ready_time: 
  secs: 10
  nsecs:         0
running_time: 
  secs: 20
  nsecs:         0
elapsed_time: 
  secs: 299
  nsecs:         0
remaining_time: 
  secs: 1
  nsecs:         0
timed_out: False
score: 20.867722087528875
---
name: "gymkhana"
state: "finished"
ready_time: 
  secs: 10
  nsecs:         0
running_time: 
  secs: 20
  nsecs:         0
elapsed_time: 
  secs: 300
  nsecs:         0
remaining_time: 
  secs: 0
  nsecs:         0
timed_out: True
score: 21.29905881898461
```
I would expect to see an additional penalty of 1.3 minus something close to 0.07 and that is pretty close to what we have.  That said, I think testing has revealed this penalty is too subtle! You have to crash like crazy for it to have any noticeable effect. What do you think of increasing it to 1 meter? If you agree I will change the documentation.

caguero:
> I would expect to see an additional penalty of 1.3 minus something close to 0.07 and that is pretty close to what we have. That said, I think testing has revealed this penalty is too subtle! You have to crash like crazy for it to have any noticeable effect. What do you think of increasing it to 1 meter? If you agree I will change the documentation.

I agree that 1 meter seems more reasonable. Changed in f81c9cc.

About the unexpected  penalty value. Could you maybe print the number of obstacles hit in `gymkhana_scoring_plugin.cc:OnFinished()` to compare the number of collisions with the penalty applied?

Thanks!

caguero:
> > I would expect to see an additional penalty of 1.3 minus something close to 0.07 and that is pretty close to what we have. That said, I think testing has revealed this penalty is too subtle! You have to crash like crazy for it to have any noticeable effect. What do you think of increasing it to 1 meter? If you agree I will change the documentation.
> 
> I agree that 1 meter seems more reasonable. Changed in [f81c9cc](https://github.com/osrf/vrx/commit/f81c9cc329a0b3558550dfc8a3fcfb04fa238412).
> 
> About the unexpected penalty value. Could you maybe print the number of obstacles hit in `gymkhana_scoring_plugin.cc:OnFinished()` to compare the number of collisions with the penalty applied?
> 
> Thanks!

One thing we could do in the future is to include the number of collisions into the `/vrx/task/info` topic.

M1chaelM:
> > I would expect to see an additional penalty of 1.3 minus something close to 0.07 and that is pretty close to what we have. That said, I think testing has revealed this penalty is too subtle! You have to crash like crazy for it to have any noticeable effect. What do you think of increasing it to 1 meter? If you agree I will change the documentation.
> 
> I agree that 1 meter seems more reasonable. Changed in [f81c9cc](https://github.com/osrf/vrx/commit/f81c9cc329a0b3558550dfc8a3fcfb04fa238412).
> 
> About the unexpected penalty value. Could you maybe print the number of obstacles hit in `gymkhana_scoring_plugin.cc:OnFinished()` to compare the number of collisions with the penalty applied?
> 
> Thanks!

Yes, sorry---I think my note just wasn't clear, but I will check again right now.

caguero:
> > > I would expect to see an additional penalty of 1.3 minus something close to 0.07 and that is pretty close to what we have. That said, I think testing has revealed this penalty is too subtle! You have to crash like crazy for it to have any noticeable effect. What do you think of increasing it to 1 meter? If you agree I will change the documentation.
> > 
> > 
> > I agree that 1 meter seems more reasonable. Changed in [f81c9cc](https://github.com/osrf/vrx/commit/f81c9cc329a0b3558550dfc8a3fcfb04fa238412).
> > About the unexpected penalty value. Could you maybe print the number of obstacles hit in `gymkhana_scoring_plugin.cc:OnFinished()` to compare the number of collisions with the penalty applied?
> > Thanks!
> 
> One thing we could do in the future is to include the number of collisions into the `/vrx/task/info` topic.

I actually implemented in #356 . I tried hitting one time and the result was the expected.

M1chaelM (approved):
Verified the latest changes work while testing PR #356. All looks good!

PR #338, opened 2021-09-09
Simple shapes as collision elements in the 2022 placards

caguero:
This pull request replaces the collision element of the 2022 placard. Before we were using the mesh directly. We're now using a few boxes strategically placed. Usually simple shapes offer better performance than meshes while checking collisions.

placard_simple_shapes


M1chaelM (approved):
Looks good! I tested to make sure the collision looks right and I can still score.

Issue #340, opened 2021-09-09
Localization warnings and errors


bsb808:
I still get a number of warnings and an error when using the robot localization example as described here: https://github.com/osrf/vrx/wiki/tutorials-ExampleVrx

Setup:  I'm working with the master branch and running on, what I believe to be, an updated noetic docker image from dockwater.

When I run ` roslaunch wamv_gazebo localization_example.launch`...

1. navsat_transform_node terminates with the following error:

```
terminate called after throwing an instance of 'GeographicLib::GeographicErr'
  what():  Latitude -33.7228, longitude 150.674 out of legal range for UTM zone 4
[wamv/robot_localization/navsat_transform_node-4] process has died [pid 1043, exit code -6, cmd /opt/ros/noetic/lib/robot_localization/navsat_transform_node /wamv/robot_localization/gps/fix:=/wamv/sensors/gps/gps/fix __name:=navsat_transform_node __log:=/home/bsb/.ros/log/a4e586bc-1192-11ec-81cf-0242ac110002/wamv-robot_localization-navsat_transform_node-4.log].
log file: /home/bsb/.ros/log/a4e586bc-1192-11ec-81cf-0242ac110002/wamv-robot_localization-navsat_transform_node-4*.log

```

*  I have partially fixed this with PR #341
*  But  the way the `datum` rosparam is specified appears to be inconsistent with the [robot_localization documentation](http://docs.ros.org/en/jade/api/robot_localization/html/integrating_gps.html) - it is missing the explicit frame declarations.  Have we verified that these frames do not need to be specified?

*  Once I fix that issue, then I still get warnings that are note documented in the wiki tutorial https://github.com/osrf/vrx/wiki/tutorials-ExampleVrx

    *  E.g.

```
[ WARN] [1631208711.225154935, 468.467000000]: Transform from wamv/imu_wamv_link to wamv/base_link was unavailable for the time requested. Using latest instead.

[ WARN] [1631208711.253446492, 468.493000000]: Transform from wamv/gps_wamv_link to wamv/base_link was unavailable for the time requested. Using latest instead.

[ WARN] [1631208713.442757568, 470.467000000]: Transform from wamv/imu_wamv_link to wamv/base_link was unavailable for the time requested. Using latest instead.

[ WARN] [1631208713.470328905, 470.493000000]: Transform from wamv/gps_wamv_link to wamv/base_link was unavailable for the time requested. Using latest instead.
```

AND

[...]

PR #341, opened 2021-09-09
Updating datum for navsat_transform_node

bsb808:
Partially addresses issue #340 

To test - follow tutorial for demonstrating localization - https://github.com/osrf/vrx/wiki/tutorials-ExampleVrx

After starting the robot_localization launch file, check that the odometry estimates are being published in local coordinates, e.g., 
Screenshot from 2021-09-09 10-48-29

### DIFF
--- a/wamv_gazebo/launch/localization_example.launch
+++ b/wamv_gazebo/launch/localization_example.launch
@@ -51,7 +51,7 @@
     <param name="broadcast_utm_transform" value="true"/>
     <param name="wait_for_datum" value="true"/>
     <param name="use_odometry_yaw" value="true"/>
-    <rosparam param="datum">[21.30996, -157.8901]</rosparam>
+    <rosparam param="datum">[-33.73, 150.67]</rosparam>
     <param name="yaw_offset" value="0"/>
     <param name="publish_filtered_gps" value="true"/>
     <remap to="/wamv/sensors/gps/gps/fix" from="/wamv/robot_localization/gps/fix" />


wamv_gazebo/launch/localization_example.launch
```diff
@@ -51,7 +51,7 @@
     <param name="broadcast_utm_transform" value="true"/>
     <param name="wait_for_datum" value="true"/>
     <param name="use_odometry_yaw" value="true"/>
-    <rosparam param="datum">[21.30996, -157.8901]</rosparam>
+    <rosparam param="datum">[-33.73, 150.67]</rosparam>
```
caguero, line 54: Should we use all the decimals for not introducing extra errors? I'm not very familiar with GCS but just checking. The values in `sydnbeyregatta.xacro` are:

```
<latitude_deg>-33.724223</latitude_deg>
 <longitude_deg>150.679736</longitude_deg>
```

bsb808: > Should we use all the decimals for not introducing extra errors? I'm not very familiar with GCS but just checking. The values in `sydnbeyregatta.xacro` are:
> 
> ```
> <latitude_deg>-33.724223</latitude_deg>
> <longitude_deg>150.679736</longitude_deg>
> ```

These are two separate datum/origin values.  We could make them the same, as a convention, but we do not need to.

I think there are at least three independent datum/origin values in the example.

1. The [`spherical_coordinates` SDF tag](http://sdformat.org/spec?elem=world) in the World file specifies the origin of the Gazebo coordinate system.  I assume that the simulation physics works in local coordinates and uses a library to transform from the local Gazebo coordinate frame to the spherical frame using this arbitrary origin value.  
I would assume that this is used for the [Gazebo GPS sensor](http://osrf-distributions.s3.amazonaws.com/gazebo/api/11.0.0/classgazebo_1_1sensors_1_1GpsSensor.html) which we don't use.

2. The  reference Latitude, Longitude, Heading and Altitude used by the [Hector GazeboRosGPS plugin](http://wiki.ros.org/hector_gazebo_plugins#GazeboRosGps).  We use this plugin and have the default values in `wamv_gps.xacro` equivalent to 1.  `<xacro:macro name="wamv_gps" params="name:=gps x:=0.3 y:=0 z:=1.3 R:=0 P:=0 Y:=0 lat:=-33.724223 lon:=150.679736 elev:=0.0 update_rate:=15">`

3. The ROS `datum` parameter for the navsat_transform_node as part of robot_localization.  This is completely independent of the simulation - users should be able to use the same robot_localization setup for sim. and for experiments.    The datum is a user specified value that defines the origin of the odometry coordinate frame.  In practice, I usually set this to the south - west of the operating area so that the local x/y values are of reasonable size (on the order of 10-1000 m) and positive.  

My recommendations are...

* Let's try to get rid of 1.  I don't think we use it - and if we don't, it causes confusion as to why we are specifying it.
* Instead of burying the lat/lon values in the `wamv_gps.xacro`, make the reference lat/lon an argument to the `wamv_gazebo.urdf.xacro` file so that this is explicitly  set in the launch files,  makes it easier to use the same macros in other areas.
* Leave 2 and 3 independent.   I worry that if we make them exactly the same, users will think that they are related and must be the same.  Also, I suspect that the hector plugin and robot_localization are using different geodetic libraries, which means that their local coordinates will be slightly different, even if the origins are the same.

Sorry for the long reply to a small PR, but if the above make sense, I'll expand the PR.

caguero: Thanks for the clarification! About (1) we use it in several places (`vrx_gazebo/src/stationkeeping_scoring_plugin.cc`, `vrx_gazebo/src/perception_scoring_plugin.cc`, `vrx_gazebo/src/wildlife_scoring_plugin.cc` and `vrx_gazebo/src/wayfinding_scoring_plugin.cc`. 

As an example, in the wildlife scoring plugin, we publish the poses of each animal in spherical coordinates. We use the `SphericalCoordinates` instance contained in the world that is initialized with the references specified in our world.

caguero: About (2): Looking at the `Hector GazeboRosGPS` code, they have a parameter `useWorldSphericalCoordinates`. See [here](https://github.com/tu-darmstadt-ros-pkg/hector_gazebo/blob/melodic-devel/hector_gazebo_plugins/src/gazebo_ros_gps.cpp#L101). 

Maybe, we can get rid of the references in our macro if we set this flag to `true`.

caguero (approved):
This change looks reasonable in any case.

Issue #344, opened 2021-09-09
Missing files for ball shooter model


bsb808:
When I run through https://github.com/osrf/vrx/wiki/tutorials-ExampleVrx

In the RVIZ launch I get errors.

First error seemed to be a .dae XML syntax issue - see PR https://github.com/osrf/vrx/pull/345

Once that is fixed, I get error of missing texture image file:

```
[ERROR] [1631215186.559655484, 581.655000000]: Error retrieving file [file:///home/bsb/vrx_ws/src/vrx/vrx_gazebo/models/ball_shooter/meshes/Glass_BaseColor.png]: Couldn't open file /home/bsb/vrx_ws/src/vrx/vrx_gazebo/models/ball_shooter/meshes/Glass_BaseColor.png
[ERROR] [1631215186.559782965, 581.655000000]: Error retrieving file [file:///home/bsb/vrx_ws/src/vrx/vrx_gazebo/models/ball_shooter/meshes/Glass_BaseColor.png]: Couldn't open file /home/bsb/vrx_ws/src/vrx/vrx_gazebo/models/ball_shooter/meshes/Glass_BaseColor.png
[ERROR] [1631215186.559884075, 581.655000000]: Error retrieving file [file:///home/bsb/vrx_ws/src/vrx/vrx_gazebo/models/ball_shooter/meshes/Launcher_BaseColor.png]: Couldn't open file /home/bsb/vrx_ws/src/vrx/vrx_gazebo/models/ball_shooter/meshes/Launcher_BaseColor.png
```

PR #352, opened 2021-09-10
Fix ball launcher in Rviz

caguero:
See issue #344.

This is a bit of  a hack but adding symbolic links to the textures from the mesh directory seems to fix the issue.

rviz_shooter


bsb808 (approved):
Works for me.

PR #354, opened 2021-09-10
Changing hector config to use Gazebo world SDF

bsb808:
This came up in PR #341 .

This PR makes use of the  the Hector `GazeboRosGPS`  a parameter `useWorldSphericalCoordinates` so that the same geographic reference is used by Gazebo and by the GPS plugin.

Here is what I did to verify:

Start with USV at gazebo origin:

```
roslaunch vrx_gazebo sydneyregatta.launch x:=0 y:=0 Y:=0
```

* Verify that at gazebo origin, GPS plugin provides lat/lon very near the specified spherical_coordinates lat/lon.

* Manually move WAM-V to roughly Gazebo x = 100, y = 0, verify that the Hector plugin reports that the latitude is roughly the same and that the longitude increases, as expected.

* Repeat for each quadrant.


vrx_gazebo/worlds/sydneyregatta.xacro
```diff
@@ -11,15 +11,15 @@
...
+    <!-- Estimated latitude/longitude of sydneyregatta
+	 from satellite imagery -->
     <spherical_coordinates>
       <surface_model>EARTH_WGS84</surface_model>
       <world_frame_orientation>ENU</world_frame_orientation>
       <latitude_deg>-33.724223</latitude_deg>
       <longitude_deg>150.679736</longitude_deg>
       <elevation>0.0</elevation>
-      <!-- For legacy gazebo reasons, need to rotate -->
-      <heading_deg>180</heading_deg>
```
bsb808, line 22: This was done to have the Gazebo x/y and Hector Lat/Lon be consistent.   I'm not sure if this has other effects?

bsb808: As @caguero mentions in PR #341 - we need to be careful with `referenceHeading`, so would appreciate a double check.

bsb808:
Also as @caguero points out in #341, we use the Gazebo spherical_coordiantes parameters in several places: e.g., vrx_gazebo/src/stationkeeping_scoring_plugin.cc, vrx_gazebo/src/perception_scoring_plugin.cc, vrx_gazebo/src/wildlife_scoring_plugin.cc and vrx_gazebo/src/wayfinding_scoring_plugin.cc.  The only change is the `referenceHeading` parameter.  It would be ideal to verify that changing that parameter doesn't have adverse effects on the scoring.

M1chaelM (changes requested):
I wasn't sure how to see the gps plugin data when running `roslaunch vrx_gazebo sydneyregatta.launch`, so instead I ran:
```
roslaunch vrx_gazebo vrx.launch x:=0 y:=0 Y:=0
```
Then I compared the Spherical Coordinates given in the Gazebo GUI with the output of
```
rostopic echo /wamv/sensors/gps/gps/fix
```
I verified that these are very close matches at the origin and in all four quadrants by running:
```
gz model -m wamv -x 500 -y 0 -z 0 -Y 0
gz model -m wamv -x -500 -y 0 -z 0 -Y 0
```
* Moving x to 500 corresponded to an increase in longitude of about  0.005, and moving to -500 corresponded to a decrease of about the same amount while the latitude remained fixed. 
* Likewise moving y to 500 corresponded to an increase in latitude of about  0.005, and moving to -500 corresponded to an decrease of about the same amount while longitude stayed fixed.
* So, these look good.

@caguero raised a concern that this update would require us to fix the positions of many objects we use in our tasks, so I ran through all six to check. It looks like the only ones that are broken are those for which we specified locations using spherical coordinates: i.e. the station keeping and wayfinding task. In these tasks the markers are not in their expected place and I can't find them anywhere in the arena, so I think this will need to also be fixed as part of this PR. All the other tasks seem to be working as expected.

I'm worried that I'm not testing this correctly because @bsb808 mentioned moving course items to new local coordinates. I'm seeing these in their expected locations already.

Screenshot from 2021-09-10 16-30-33

Am I missing something?

bsb808:
@M1chaelM See PR #358

M1chaelM:
Since this is part of the same work as PR #358 and both are relatively small code changes and they don't overlap I merged that branch into this one so they can be reviewed together. I hope I'm not failing to account for anything, but I left the other PR open for now just in case.

M1chaelM (changes requested):
It looks like the latest changes do in fact require repositioning of all models, as expected. I've run through our various launch files to make a quick inventory of what needs to be done before we can merge this:
1. stationkeeping: The goal is correct. Only the wamv needs to be relocated.
2. wayfinding: same as stationkeeping
3. perception: very odd. The wamv spawns under the regatta ground, and some of the objects appear in front of it, but others appear under different ground far away.
4. wildlife: wamv and animals are together under the regatta ground, so all will need to be moved.
5. gymkhana: course, wamv and pinger location are all incorrect
6. perception: wamv, light buoy and dock are all incorrect

## Conclusions:
* Of the 6 items above, at least a few should be simple fixes, but perception seems more daunting.
* It seems to make sense to fix them in this PR before merging rather than merge and leave many aspects of the competition in an unusable state.
* I think we will also want to review the tasks more thoroughly than I have above to make sure scoring is working correctly.

## Proposal 
Given the magnitude of the task and the lateness of the hour, I'd be in favor of making this PR part of the 1.5.1 patch release to give us a chance to test everything thoroughly. Although it's obviously important, I don't think it's a blocker for 1.5 since the simulation is still very functional without it. I can put a note in the changelog that this is a known issue and will be fixed in the next release.

@bsb808, @caguero: thoughts?

bsb808:
Sounds like a plan to me.

bsb808:
Also, see #358 - has additional documentation on this issue.

M1chaelM:
Some documentation for posterity (i.e. just in case I miss some models that need to be fixed later):
* Our height map has rotated almost 180 degrees, but not quite. I think this is because we were using 3.14 to approximate the rotation. Because of this, adjustments to model coordinates are not perfectly clean.
* Despite the above, we can get very close to the correct new position by changing the sign of both the x and y coordinates and rotating by 180 degrees.
* Sometimes the above transformation is good enough.
* Even if it isn't, it gets you to a better starting point from which to make further adjustments.

M1chaelM:
Todo:
- [ ] VRX launch: move course elements in example_world
- [x] station keeping: move wamv and rotate goal
- [x] wayfinding: move wamv
- [x] perception: move wamv and check buoy locations
- [ ] wildlife: move wamv and animal locations, check animal motion
- [ ] gymkhana: move wamv, course and pinger
- [ ] scan_dock_deliver: move wamv and course elements

M1chaelM:
I think this PR is now in a state where changes should be mostly unnoticeable to the user (except that the orientation is now correct). To test, I'd like to re-run through the tutorials and verify that everything looks the same and the scoring is right. I've taken a pass at this but it would be good to have an independent check. Could everyone please run through the tutorials you're tagged on below and check the box to confirm they work as expected? (If you run into a problem just leave a comment describing it; don't worry about fixing for now.)
 
@caguero :
- [ ] https://github.com/osrf/vrx/wiki/tutorials-ExampleVrx
- [ ] https://github.com/osrf/vrx/wiki/vrx_2022-wildlife_task
- [ ] https://github.com/osrf/vrx/wiki/vrx_2022-scan_dock_deliver_task

@crvogt 
- [ ] https://github.com/osrf/vrx/wiki/vrx_2022-station_keeping_task
- [ ] https://github.com/osrf/vrx/wiki/vrx_2022-wayfinding_task
- [ ] https://github.com/osrf/vrx/wiki/vrx_2022-perception_task

@bsb808
- [ ] https://github.com/osrf/vrx/wiki/tutorials-PropulsionConfiguration

@j-herman
- [ ] https://github.com/osrf/vrx/wiki/vrx_2022-gymkhana_task

Thanks all for the help with this!

j-herman:
Confirmed gymkhana task still looks and works the same, including scoring.

crvogt:
Station keeping, wayfinding, and perception looking and working the same!

caguero:
My worlds all look good! I made a minor update in the `gz model` command to be used in the "Scan, dock and deliver" world. 7a3799c

M1chaelM (approved):
Looks like everything is now ready to go. Thanks all!

PR #356, opened 2021-09-10
Adding number of collisions to the task message

caguero:
I moved the logic to account for the collisions to the base scoring plugin and include it in the task message.

### DIFF
--- a/vrx_gazebo/include/vrx_gazebo/gymkhana_scoring_plugin.hh
+++ b/vrx_gazebo/include/vrx_gazebo/gymkhana_scoring_plugin.hh
@@ -46,9 +46,6 @@ class GymkhanaScoringPlugin : public ScoringPlugin
   protected: void BlackboxCallback(const vrx_gazebo::Task::ConstPtr& msg);
 
   // Documentation inherited.
-  private: void OnCollision() override;
-
-  // Documentation inherited.
   private: void OnFinished() override;
 
   /// \brief Pointer to the update event connection.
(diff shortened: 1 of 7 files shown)


M1chaelM (approved):
Tested by running:
```
roslaunch vrx_gazebo gymkhana.launch extra_gazebo_args:=--verbose
```
and
```
rostopic echo /vrx/task/info
```
Then navigating the course and hitting some obstacles. The `num_collisions` counter is working as expected. I notice that there is a discrepancy between the number of collision messages produced and the number of collisions counted, but I think this is good. It looks like I get a collision message for every single bump, even if several of them happen as part of what an onlooker would probably think of as a single incident, and multiple collision messages can have the same collision number. The counter only increments for each separate incident. This probably also explains why I didn't lose as many points as I thought I would in previous testing.

After five collisions, I waited for the task to end and got the following final messages from `/vrx/task/info`:
```
name: "gymkhana"
state: "running"
ready_time: 
  secs: 10
  nsecs:         0
running_time: 
  secs: 20
  nsecs:         0
elapsed_time: 
  secs: 299
  nsecs:         0
remaining_time: 
  secs: 1
  nsecs:         0
timed_out: False
num_collisions: 5
score: 35.41200638594324
---
name: "gymkhana"
state: "finished"
ready_time: 
  secs: 10
  nsecs:         0
running_time: 
  secs: 20
  nsecs:         0
elapsed_time: 
  secs: 300
  nsecs:         0
remaining_time: 
  secs: 0
  nsecs:         0
timed_out: True
num_collisions: 5
score: 40.294599302037504
```
These are exactly as I would expect, so I think this is ready to go.

multiple contacts with a single obstacle that are all part of what I would describe as a single collision.

caguero:
There's a parameter `<collision_buffer>` that sets the minimum time between collisions.

Issue #353, opened 2021-09-10
Is the Sydney Regatta map orientation correct?


bsb808:
I may be confused, but I think the Sydney Regatta map is incorrectly oriented.  

I launch the regatta world with the USV at the origin...

```
roslaunch vrx_gazebo sydneyregatta.launch x:=0 y:=0 Y:=0
```

Here is what I believe is the birds eye view of the location with North up.
![Screenshot from 2021-09-10 11-03-52](https://user-images.githubusercontent.com/8921143/132898258-9ce26e00-746d-470b-b31e-ba19ecbc40e4.png)

And here is the google map view...
![Screenshot from 2021-09-10 11-05-04](https://user-images.githubusercontent.com/8921143/132898283-fa7fe2ef-14f6-48fb-8a5d-e9863588b377.png)

Are we 180 degrees off?  

Is this as simple as applying a Yaw rotation to the model origin?

M1chaelM:
Uh oh, I wonder if this is related to:
* #312 

@crvogt noticed last month that the GPS coordinates were reversed. The PR is not referenced in the issue, but I will look for it to see whether this is a side effect of our solution.

M1chaelM:
Update: #312 was resolved with:
* #313
Which rotated the arena 180 degrees in this commit: https://github.com/osrf/vrx/pull/313/commits/b574656fe9a2e4f6ab6e3f99485e83a6f095c950

I could be misunderstanding, but it seems like this fix had the unintended consequence you are noticing here.

bsb808:
> Update: #312 was resolved with:
> 
> * [Update Station Keeping Task #313](https://github.com/osrf/vrx/pull/313)
>   Which rotated the arena 180 degrees in this commit: [b574656](https://github.com/osrf/vrx/commit/b574656fe9a2e4f6ab6e3f99485e83a6f095c950)
> 
> I could be misunderstanding, but it seems like this fix had the unintended consequence you are noticing here.

The first thing to make sure we have correct is the Gazebo coordinates (ENU) and how they relate to the Sydney map.  The spherical_coordinates parameters should be independent of this .  I don't believe that changing the <heading_deg> parameter will affect the gazebo coordinate orientation.

caguero:
I see a 180 deg. rotation applied in the world [here](https://github.com/osrf/vrx/blob/master/vrx_gazebo/worlds/sydneyregatta.xacro#L41).

caguero:
[...]

PR #358, opened 2021-09-11
Toward consistent local (Gazebo) and geographic (GPS) coordinates.

bsb808:
Begins to address Issue #353 .

This PR is on top of PR #354 , so probably best to merge that one first, then rebase before merging.

* Remove the 180 deg rotation of the Sydney Regatta which causes issue #353.
* Modification the model pose values in`sydneyregatta.xacro` so that the models (post, ground stations, etc.) are in the same location relative to the Regatta map.
* Mod. of world file to change camera view
* Mod. of the launch file change WAM-V default initial pose.

I checked the consistency of the Gazebo / Hector GPS coordinates with Google Earth...

sydney_regatta

I have not yet tried the various VRX tasks to see what other local coordinate values need to be modified to maintain the desired relative location of models.


M1chaelM:
@bsb808 Looking at this now. Since it is on top of #354, why don't we change the merge target to `bsb/use_spherical` so we can test both together before merging either into the master?

caguero:
Consider to look at [this pull request](https://github.com/ignitionrobotics/ign-math/pull/235) in Ignition Math in case we're impacted with the `SphericalCoordinates` fix detailed in that PR.

bsb808:
Also, see https://github.com/osrf/vrx/pull/358 - has additional documentation on this issue.

bsb808:
Closing b/c we are working directly on #354

Issue #279, opened 2021-07-22
Include RobotX 2022 elements when launching vrx.launch


caguero:
In the past, when launching `sandisland.launch` the simulation loaded an empty Sand Island world with a very basic WAM-V without sensors. The old `vrx.launch` loaded the WAM-V in the competition configuration (with all the sensors) and a Sand Island world with multiple RobotX elements to practice.

Right now, launching `sydneyregatta.launch` has a consistent behavior as it loads a basic WAM-V and an empty Sydney Regatta world. However, launching `vrx.launch` loads the exact same world as `sydneyregatta.launch` without any RobotX elements. We need to update the code to include some of the updated elements (docks, navigation channel, etc.).

M1chaelM:
@bsb808 Could you take a look at this?

bsb808:
@M1chaelM I will try, but I don't know if I'll have time by COB tomorrow.  I spent most of the day on issues and PRs a related to the video demos and have a pretty hectic day tomorrow.   I'll try to take a look later tonight.

M1chaelM:
@bsb808 OK, thanks! If you can't get to it let me know and we'll figure something out.

M1chaelM:
@bsb808 I can work on this unless you've already started on it.

bsb808:
@M1chaelM Cool.  I haven't started.   This will likely need to be re-worked with PR #358

PR #359, opened 2021-09-11
adding 2022 elements to vrx.launch

M1chaelM:
This resolves issue #279

To test, run:
```
```
You should see the light buoy, docks, gymkhana course and our 3 animals:

vrx_2022_launch

### DIFF
--- a/vrx_gazebo/launch/vrx.launch
+++ b/vrx_gazebo/launch/vrx.launch
@@ -15,6 +15,9 @@
   <arg name="wamv_locked" default="false" />
   <!-- Start paused? -->
   <arg name="paused" default="false"/>
+  <!-- Acoustic pinger position(s) -->
+  <arg name="pinger_params" default="$(find vrx_gazebo)/config/pinger.yaml"/>
+
   <!-- Initial USV location and attitude-->
   <arg name="x" default="532" />
   <arg name="y" default="-162" />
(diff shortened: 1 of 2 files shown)


bsb808:
I think we need to first address PRs #354 and #358  - then when those are merged, rebase this branch to see if it still works as expected.  I suspect that the model poses will need to be modified.

M1chaelM:
> I think we need to first address PRs #354 and #358 - then when those are merged, rebase this branch to see if it still works as expected. I suspect that the model poses will need to be modified.

@bsb808  Sure, I can take a look at those now. That said, I have already created this file and it appears to be working, so I think it would be safe to go ahead and approve. If changes are needed I will add them as part of the PR that introduces the need for new poses. Is there any downside to this?

bsb808:
> > I think we need to first address PRs #354 and #358 - then when those are merged, rebase this branch to see if it still works as expected. I suspect that the model poses will need to be modified.
> 
> @bsb808 Sure, I can take a look at those now. That said, I have already created this file and it appears to be working, so I think it would be safe to go ahead and approve. If changes are needed I will add them as part of the PR that introduces the need for new poses. Is there any downside to this?

Just seems like a little more work to test this modification twice, once now and then again after we fix the coordinates.

bsb808 (approved):
Good to go for 1.5

PR #361, opened 2021-09-12
Enable reflections and refractions

caguero:
The reflection/refraction effect wasn't working. This pull request solves the problem and enables back this effect.

reflections


M1chaelM (approved):
Works as advertised. Thanks!

PR #362, opened 2021-09-12
Fix pinger.launch

caguero:
While testing the [acoustic pinger tutorial](https://github.com/osrf/vrx/wiki/acoustic_pinger_tutorial), the following command failed:

```
 roslaunch vrx_gazebo pinger.launch
```

with this error message:

```
ERROR: cannot launch node of type [vrx_gazebo/pinger_visualisation.py]: Cannot locate node of type [pinger_visualisation.py] in package [vrx_gazebo]. Make sure file exists in package path and permission is set to executable (chmod +x)
```

The problem is that `pinger.launch` contains a typo. This pull request fixes it.


M1chaelM (approved):
Works for me now. Thanks for catching this. I changed the spelling in the node name to also use the American version, just for consistency.

Issue #298, opened 2021-07-29
Organize launch and world files to make it possible to launch 2019 and 2022 tasks easily.


caguero:
Ideas: Pass an extra parameter (year) and that will dictate the initial pose of the WAM-V and which world to launch. Also, maybe it makes sense to create subdirectories for the worlds.

M1chaelM:
Resolved by #368

PR #368, opened 2021-09-23
Move 2019 assets to vrx_2019

caguero:
See #298.

This pull request moves all the 2019 world and launch files to a new `vrx_2019` package. I also restored some of the 2019 world files that were updated to the 2022 rules.

How to test it:

* I suggest removing the `build` and `devel` directories and recompile.
* Then, you can try launching all the worlds in the new package. Example:

```
roslaunch vrx_2019 vrx.launch
```

and confirm that the world loads correctly.


M1chaelM (approved):
To test, I ran through our [tutorials for the 2019 tasks](https://github.com/osrf/vrx/wiki/vrx_2019-task_tutorials)  and tested all 6 tasks. Everything worked. I made updates to the Wiki as follows:
* Replaced `vrx_gazebo` with `vrx_2019` for all task launch commands.
* Updated the `/vrx/station_keeping/rms_error` topic to the new `/vrx/station_keeping/mean_pose_error`.
* Added notes to point out the station keeping and wayfinding scoring functions have changed from the original competition.
I think this is fine. The only thing it breaks is if there is some good reason someone would want to go back and try to reproduce the same scores or something. In that case, they'll have to do a little more work. Meanwhile, I think the main use case is to have extra practice arenas available.

PR #369, opened 2021-09-24
Practice worlds 2022

crvogt:
Added new practice worlds. Each task has three worlds in the `2022_practice`  directory. New folders have been added for the navigation and obstacle course portions of the gymkhana task. 

The navigation and obstacle courses still need to be randomized for the gymkhana trials.


M1chaelM:
I made some small changes to the station keeping practice worlds:
* The goal pose orientation is now different for all 3 worlds.
* Parameters for world 1 are slightly modified so it is more in the "middle" between 0 and 2.

M1chaelM:
@crvogt I just noticed the wildlife encounter worlds are missing from our set. Could you add these?

M1chaelM:
I've made some adjustments to the wayfinding worlds. I think these should be ready (but would appreciate a double-check). Also, there are a few things that I noticed that are probably worth making a note of for future wayfinding world generation:
* Our goal pose and waypoint markers are direction sensitive. They were initially all set in the same direction, which isn't necessarily wrong but it can make the course challenging, so just wanted to make sure you're aware of it.
* The environmental parameters need to stay within the min/max we specify on page 11 of the technical guide:  https://robonation.org/app/uploads/sites/2/2021/09/VirtualRobotX2022_Technical-Guide_v1.0.pdf. We are mostly doing this but for example wayfinding1.world sets the wind mean velocity to 20, which is out of bounds (and very challenging--I personally can't get the USV off the beach under these conditions).
* This might be clear already, but just in case: we generally try to organize the examples so 0 is easy and uses the low end of the environmental envelop, 1 is intermediate and uses parameters in the middle, and 2 is hard and uses parameters near the upper end. 
* The visual waypoint markers are for debugging only--the wamv won't be able to "see" them in the real competition. Their coordinates and heading are provided through a ros topic. The consequence of this is that fog doesn't really affect this task too much (though there's nothing wrong with adding it imo).

caguero (changes requested):
I've tried the wildlife and gymkhana worlds. Here are a few comments:

* I don't think I see changes in fog or ambient light.
* Are you thinking on moving the buoys and gates in the gymkhana world?

crvogt:
> I've tried the wildlife and gymkhana worlds. Here are a few comments:
> 
> * I don't think I see changes in fog or ambient light.
> * Are you thinking on moving the buoys and gates in the gymkhana world?

I spoke with Michael, and regarding fog I removed it from all but the perception task. There was concern regarding the tree appearance when fog was added.

I hadn't considered changing the ambient light, I'll check the 2019 worlds and update.

Ran out of time last night regarding the buoys and gates, working on that right now.

crvogt:
@caguero Added and adjusted discussed parameters. Largely took some navigation cues from the 2019 practice worlds, though reduced the number of gates/lightly simplified.

caguero (changes requested):
I also tried the scan, dock and deliver worlds. It all looks good to me, except for a disagreement between the color sequence in the light buoy and the placards in the dock. According to the rules, the first color of the sequence sets the color of the gate. The last color of the color sequence dictates the figure of the gate. I don't think there's a good match between the color sequence and the placards. None of the worlds have the expected placard in the dock.

caguero (changes requested):
`gimkhana1` has two pairs of `black-red` buoys. However, we should only have one `black-red` gate delimiting the last gate.

`gimkhana2` has two pairs of `white-red` buoys. However, we should only have one `white-red` gate delimiting the first gate.

All intermediate gates should be `green-red`.

caguero (approved):

vrx_gazebo/launch/perception_task.launch
```diff
@@ -2,7 +2,7 @@
 <launch>
   <env name="ROSCONSOLE_CONFIG_FILE" value="$(find vrx_gazebo)/config/custom_rosconsole.conf"/>
   <!-- Gazebo world to load -->
-  <arg name="world" default="$(find vrx_gazebo)/worlds/perception_task.world" />
+  <arg name="world" default="$(find vrx_gazebo)/worlds/2022_practice/gymkhana2.world" />
```
caguero, line 5: Is this change on purpose?

caguero (changes requested):

vrx_gazebo/launch/perception_task.launch
caguero: Fixed in c6eada1.

PR #371, opened 2021-10-05
Deconflicting plugin name with VORC

bsb808:
The name Gymkhana plugin name was already in use by VORC, so introducing the same name in VRX introduces a build error.  This changes the plugin library name, leaving the source file names that same.

Currently working on testing with VORC - would appreciate someone testing with the VRX Gymkhana task scoring.


bsb808:
I verified that the base case for VRX works - roslaunch vrx_gazebo vrx.launch

caguero (approved):
It looks good to me!

PR #374, opened 2021-10-05
restore style checks

M1chaelM:
This PR updates cpplint.py so it will run on systems that only have python3, and enables it on those systems. To review, a scan through changes to cpplint.py and code_check.sh should be sufficient. The changes to other files are to correct style issues found by cpplint.py and the automatic tests should make sure the code is passing these checks.


M1chaelM:
As a side note, some of the python3 compatibility changes were inspired by this PR to the Google styleguide: https://github.com/google/styleguide/pull/349/files

caguero (approved):
Looks good to me!

Issue #367, opened 2021-09-22
VRX Cloudsim real time factor < 0.7


M1chaelM:
VRX builds and runs correctly on our cloudsim image. However, the real time factor is low. Even running locally, I see RTF < 0.7:
![Screenshot from 2021-09-22 09-47-29](https://user-images.githubusercontent.com/8611855/134389300-e3a806c1-337b-45a7-a178-ec0d3e276600.png)

M1chaelM:
This could potentially be a problem with the Nvidia driver. The graphical "Additional Drivers" tab does not seem to be working, but command line tools report that the image is using the Nvidia driver:

![Screenshot from 2021-09-22 10-11-53](https://user-images.githubusercontent.com/8611855/134390549-b5227ea9-d829-4140-b811-4946b9158e9a.png)

This is the same driver as the host system, so I'm not yet sure if this is the problem. Since this image is relatively old now I may try rebuilding to see whether the issue persists with updated software.

M1chaelM:
@bsb808 When you try running locally could you let me know whether you are also still seeing a slow RTF?

bsb808:
@M1chaelM - I can try to repeat what you are seeing, but it sounds like we are seeing consistent low performance both running on AWS and running locally.  This seems to suggest a problem with the image.   Have you tried this with both the matlab and non-matlab image?

M1chaelM:
To troubleshoot this I compared 4 different scenarios:
* Running our cloudsim image locally and accessing through the browser using novnc
* Running our cloudsim image locally without novnc, using x11 directly as we do with our dockwater image
* Running our dockwater image (locally with x11)
* Running our dockwater image through novnc, using `rocker --nvidia --novnc --turbovnc --user --user-override-name=developer noetic:latest`
* Running on the "bare metal" using a host install

Since the default vrx.launch demo has a lot of floating objects, including 3 moving animals, I also ran each scenario using the simpler station_keeping.launch file. 

|   | cloudsim novnc local | cloudsim x11 | dockwater novnc local | dockwater x11   | bare metal  |
|---|---|---|---|---|---|
|vrx.launch   | 0.59  | 0.75  | 0.59 | 0.98 | 0.98  |
| station_keeping.launch  |  0.59 | 0.86  | 0.59 | 0.98   | 0.98  |

## Takeaways
* Using NoVNC seems to produce a major performance hit.
[...]

PR #380, opened 2021-10-15
Improve Sydney Regatta performance

caguero:
See issue #367.

This pull request adds a new simplified mesh for the Sydney Regatta world to be used for collision checking. I can see better CPU usage in my machine and it's running in real-time.

Maybe @bsb808 or @M1chaelM  can test it on the cloud to check if the speedup is more evident.

@JonathanWheare, feel free to give it a shot.


caguero:
> @caguero I'm still seeing a RTF of 0.65 in the cloud using this collision model. This is better than the 0.4 Brian was seeing, though. When I totally disable the collision for the Sydney Regatta model, RTF bumps up to 1.0, though FPS is still around 2.3. If I also delete the visual model, I go up to RTF 1.0 and FPS 6-7, so it does seem that this model is having a big impact.
> 
> Should I test anything else? I think this is a step in the right direction. Not sure if there is more to change before we merge.

Thanks for testing in the cloud. Let me see if we can improve things before merging.

caguero:
> > @M1chaelM , I updated the mesh. Do you mind to give it a shot again in the cloud to see where we are?
> 
> It seems to have gotten a bit worse for some reason. I'm seeing RTF 0.59 and FPS 3.23.

Thanks, let me keep looking into it...

M1chaelM:
> Can you confirm it on your end?
We are waiting on a fix for cloudsim. We've been unable to launch new machines to test this for a couple of days (apparently due to a shortage of EC2 instances!?). I will test as soon as we can get something to try it on.

M1chaelM:
@caguero I was able to test this today and I'm now getting a RTF between 0.91-0.95, so much better. The frame rate is still extremely low at around 2.75. This varies depending on what's in the scene, but it's never higher than 9. Deleting the arena height map raises FPS to around 9 also.

caguero:
> @caguero I was able to test this today and I'm now getting a RTF between 0.91-0.95, so much better. The frame rate is still extremely low at around 2.75. This varies depending on what's in the scene, but it's never higher than 9. Deleting the arena height map raises FPS to around 9 also.

Great! About the FPS, did you try disabling the tree visuals?

M1chaelM:
> Great! About the FPS, did you try disabling the tree visuals?

Surprisingly, this seems to improve both the RTF and FPS a little bit. RTF goes to 0.95-0.96 and FPS goes to 3.1-3.3.

JonathanWheare:
Using this version, I am seeing a real time factor of around 0.97 to 0.98 with 120% CPU on my laptop.  Reverting to the current master branch, I am getting about 0.6 RTF with 160-180%CPU.  Seems like a significant improvement.

caguero:
I feel this change is good enough for this pull request. How about we stop here and try to tackle the rendering performance issue in a separate patch?

M1chaelM (approved):
I agree the improvement is good enough to warrant merging this PR.

PR #383, opened 2021-10-21
fix generate_worlds capability

M1chaelM:
This PR fixes some minor errors that were preventing generate_worlds.py from working, and does some housekeeping to distinguish between world .yaml files we made for 2019 and the ones we need to make for 2022.

I also created one example stationkeeping.yaml file for 2022. Once this is merged I will create the yaml for the other tasks.

The easiest way to test is to use the `prepare_task_trials.bash` script include with vrx-docker:
```
cd ~/vrx_ws/src
git clone git@github.com:osrf/vrx-docker
cp vrx/vrx_gazebo/worlds/yamls/stationkeeping.yaml vrx-docker/task_config/
cd vrx-docker
./prepare_task_trials.bash stationkeeping
```
This should generate 3 stationkeeping worlds in `vrx-docker/generated/task_generated/stationkeeping/worlds/`. Their parameters should match what is specified in `vrx/vrx_gazebo/worlds/yamls/stationkeeping.yaml`.

If all goes well the following commands should bring up the practice worlds.
```
roslaunch vrx_gazebo vrx.launch world:=`pwd`/generated/task_generated/stationkeeping/worlds/stationkeeping0.world verbose:=true
roslaunch vrx_gazebo vrx.launch world:=`pwd`/generated/task_generated/stationkeeping/worlds/stationkeeping1.world verbose:=true
roslaunch vrx_gazebo vrx.launch world:=`pwd`/generated/task_generated/stationkeeping/worlds/stationkeeping2.world verbose:=true
```
These are designed to be very similar to the worlds in `vrx/vrx_gazebo/worlds/2022_practice/` but they include some more recent modifications to the plugins. Once this is merged we should replace the worlds in 2022_practice with the generated worlds so everything matches.

### DIFF
--- a/vrx_gazebo/worlds/yamls/stationkeeping.yaml
+++ b/vrx_gazebo/worlds/yamls/stationkeeping.yaml
@@ -1,0 +1,133 @@
+constant:
+    steps: 1
+    macros:
+        sydneyregatta_minus_scene: 
+            -
+    sequence:
+
+environment:
+    steps: 3
+    macros:
+        stationkeeping:
+            -
+        ocean_waves:
+            -
+        usv_wind_gazebo:
+            -
+        scene_macro:
+            -
+    sequence:
+        0:
+            stationkeeping:
+                - lat: "-33.7227180"
+                  lon: "150.6740310"
+                  heading: "0.00"
+                  /**wp_markers: "
+                  <markers>
+
+                  <scaling>0.2 0.2 2.0</scaling>
+
+                  <height>0.5</height>
+
+                  </markers>"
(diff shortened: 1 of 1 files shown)


M1chaelM:
Note: I'm noticing an error message when the task moves to the OnReady state:
```
[Err] [scoring_plugin.cc:237] Unable to release [wamv_external_pivot_joint]
[Err] [scoring_plugin.cc:237] Unable to release [wamv_external_riser]
```
I haven't looked into it closely but it appears also when running with the `2022_practice` worlds so I think it may warrant a separate issue.

PR #386, opened 2021-10-28
Add ball shooter to components

crvogt:
Related to #373 

Ball shooter can now be configured in the same way as sensors and follows the same compliances.

May test by generating a new WAM-V similar to [this tutorial](https://github.com/osrf/vrx/wiki/tutorials-Creating%20a%20custom%20WAM-V%20Thruster%20and%20Sensor%20Configuration%20For%20Competition)

- Create a `my_wamv` directory. 
- Copy `example_thruster_config.yaml` and `example_component_config.yaml` into the directory.
- To generate the WAM-V urdf, run
 ```roslaunch vrx_gazebo generate_wamv.launch thruster_yaml:=$HOME/my_wamv/example_thruster_config.yaml component_yaml:=$HOME/my_wamv/example_component_config.yaml wamv_target:=$HOME/my_wamv/my_wamv.urdf```
- To test with the new WAM-V, run `roslaunch vrx_gazebo sydneyregatta.launch urdf:=$HOME/my_wamv/my_wamv.urdf`

PR #387, opened 2021-11-02
Remove p3d from sensor config example

caguero:
We don't allow a `p3d` sensor (ground truth) as part of the `sensor_config.yaml` anymore. However the example that we're shipping includes a `p3d` sensor.


caguero:
Never mind, this change is included in pull request #386 .

PR #392, opened 2021-11-05
Add a xacro for generating gymkhana worlds

M1chaelM:
Adds gymkhana.xacro, for use with the `generate_worlds` script to create practice worlds. This PR should be tested in conjunction with [PR 41](https://github.com/osrf/vrx-docker/pull/41) in the `vrx-docker` repo, using the testing instructions listed there.

### DIFF
--- a/vrx_gazebo/worlds/xacros/gymkhana.xacro
+++ b/vrx_gazebo/worlds/xacros/gymkhana.xacro
@@ -30,7 +30,7 @@
       <contact_debug_topic>/vrx/debug/contact</contact_debug_topic>
       <initial_state_duration>10</initial_state_duration>
       <ready_state_duration>10</ready_state_duration>
-      <running_state_duration>300</running_state_duration>
+      <running_state_duration>${running_duration}</running_state_duration>
       <!-- Per-plugin flag, different from env var VRX_EXIT_ON_COMPLETION.
            Respect top-level plugin finished status. -->
       <per_plugin_exit_on_completion>true</per_plugin_exit_on_completion>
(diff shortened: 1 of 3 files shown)


vrx_gazebo/worlds/xacros/gymkhana.xacro
```diff
@@ -0,0 +1,104 @@
+<?xml version="1.0"?>
+<world xmlns:xacro="http://ros.org/wiki/xacro">
+  <xacro:macro name="gymkhana" params="name='gymkhana' 
+   nav_uri:=short_navigation_course0
+   obstacle_uri:=obstacle_course
+   nx:=-524 ny:=186 nz:=0 nR:=0 nP:=0 nY:=-1.44 
+   ox:=-477 oy:=275 oz:=0 oR:=0 oP:=0 oY:=2.54 
+   gx:=-483 gy:=295.5 
+   running_duration:=300 **gates">
```
caguero, line 9: Is this parameter used?

M1chaelM: Which one? running_duration or gates?

M1chaelM: running_duration is not currently being used but could be useful to change during testing. Likewise, gates has the same value for everything right now because that is what matches the practice worlds we designed so far, but I could imagine varying the values in the future.

M1chaelM: Ah! I see what you mean. I forgot to put the variable in. Will fix that now.

caguero: It's OK to not use it and fallback on the default value but what I mean is that `running_duration` should appear in all the plugins as `<running_state_duration>${running_duration}</running_state_duration>`, right?

M1chaelM: Yes! I think our messages are missing each other slightly but I understand and just fixed this.

caguero (approved):

PR #394, opened 2021-11-07
Update pinger.yaml to comply with acoustic_pinger_tutorial

Tinker-Twins:
Updated the `pinger.yaml` file to include pinger position `[65.0, -47.5, -2]` in order to comply with the results produced by [Visualize sensor measurements in RViz](https://github.com/osrf/vrx/wiki/acoustic_pinger_tutorial#visualize-sensor-measurements-in-rviz) section in `acoustic_pinger_tutorial`.


caguero (changes requested):
Thanks for detecting this inconsistency in the tutorial. How about we create a solution that works for the 2019 world and the current one?

I suggest to create a new `pinger.launch` under `vrx_2019/launch` pointing to a config file under `vrx_2019/config/pinger.yaml`? Then, you can set the new `yaml` file with the old values `[65.0, -47.5, -2]`.

We should be able to launch both alternatives with:
```
roslaunch vrx_gazebo pinger.launch
```
or
```
roslaunch vrx_2019 pinger.launch
```

caguero (changes requested):
You also need to install the `config` directory. I suggest to add this block to `vrx_2019/CMakeLists.txt`:

```
# Install all the config files
install(DIRECTORY config/
  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/config)

```

Tinker-Twins:
Made the necessary changes and tested locally by cloning the updated repository.

caguero (approved):

PR #397, opened 2021-11-10
Enable shooting the ball with the joystick

j-herman:
To test:
Run scan/dock/deliver task
roslaunch vrx_gazebo scan_dock_deliver.launch
Run teleop node
To confirm the ball shooter actually fires, listen on the appropriate topic
rostopic echo /wamv/shooters/ball_shooter/fire
Hit the left lower trigger and you should see the ball shoot, as well as an empty message (---) on the firing topic.

The code is added in twist2thrust.py, which seemed logical since that's running anyways to map between joystick and wamv commands.  It could be broken out separately if desired.


M1chaelM (approved):
Looks good! All works as advertised.

PR #400, opened 2021-11-18
Adding wind and paused to stationkeeping demonstrations

bsb808:
This PR makes some small tweaks to the launch and world files for the station-keeping task demo.  These were created as part of the kickoff demonstration to illustrate parameterized wind for testing.  Exposing the `paused` argument in the launch file is typical and expected.


M1chaelM (approved):
Looks good. Pausing works. I was confused by the change to the wind because it sets the velocity to 0, but looking at your slide I think the goal is to provide the user a template where they can change the parameter if they want.

bsb808:
> I was confused by the change to the wind because it sets the velocity to 0, but looking at your slide I think the goal is to provide the user a template where they can change the parameter if they want.

Yes.

PR #411, opened 2021-12-01
Update perception_task.launch to include correct path to perception_task.world

Tinker-Twins:
Path to the `perception_task.world` file in the `perception_task.launch` file seems to be incorrect/outdated since the `perception_task.world` file is no longer built to the `/vrx_gazebo/worlds/2022_practice/` path, but instead to the `/vrx_gazebo/worlds/` path (according to the [updated repository structure](https://github.com/osrf/vrx/blob/master/vrx_gazebo/worlds/perception_task.world.xacro)). Thus, when executed, instead of bringing up the `perception_task` world, it merely brings up `empty_world` in Gazebo simulator.

In order to resolve the aforementioned issue, `perception_task.launch` file as been updated to now include the correct path to `perception_task.world` file.


M1chaelM:
@crvogt My guess is that this path was changed accidentally in PR #369 . Could you confirm?

crvogt:
> @crvogt My guess is that this path was changed accidentally in PR #369 . Could you confirm?

Correct, looks like it was left from testing in that PR.

M1chaelM (approved):
Looks good. Thanks for catching this!

Issue #420, opened 2022-01-25
Wildlife Encounter and Avoid Task does not provide correct spherical coordinates


alexperez33:
It appears the spherical coordinates provided by the wildlife encounter and avoid task topic ```/vrx/wildlife/animals/poses``` do not correspond with the locations of the actual animals. After some investigation, it appears that the poses provided in the ```wildlife_task.world.xacro``` file seem correct. But when converted to spherical coordinates in the ```SphericalFromLocal``` function within the ```wildlife_scoring_plugin.cc```, the poses no longer seem to correspond to that of the animals.

One way of showing this is by typing in the topic's latitude and longitude into this website: https://www.latlong.net/lat-long-utm.html. Moreover, when converting the provided latitude and longitude to a pose in reference to enu, I am provided with a pose that corresponds to that of the website above, not the actual animals.

Please let me know if I am missing something.

j-herman:
Closing this issue based on merge of https://github.com/osrf/vrx/pull/428.  The topic is now publishing the correct lat/ long.

PR #421, opened 2022-01-25
Potential fix for wildlife encounter animal poses

alexperez33:
(Issue #420)
This converts the animal xyz coordinates to a global Cartesian frame before calling the ```SphericalFromLocal``` function in ```wildlife_scoring_plugin.cc```. I am unsure exactly why this works, but I properly receive the spherical coordinates from the ```vrx/wildlife/animals/poses``` topic now.

I do notice that the function I added that converts the xyz coordinates to a global Cartesian frame (```GlobalFromLocal```) simply rotates the position of the animals about the Gazebo origin frame (both x and y are negated).

### DIFF
--- a/vrx_gazebo/src/wildlife_scoring_plugin.cc
+++ b/vrx_gazebo/src/wildlife_scoring_plugin.cc
@@ -461,13 +461,13 @@ void WildlifeScoringPlugin::PublishAnimalLocations()
     // Conversion from Gazebo Cartesian coordinates to spherical.
 #if GAZEBO_MAJOR_VERSION >= 8
     const ignition::math::Pose3d pose = buoy.link->WorldPose();
-    const ignition::math::Vector3d position = 
+    const ignition::math::Vector3d position =
       this->world->SphericalCoords()->GlobalFromLocal(pose.Pos());
     const ignition::math::Vector3d latlon =
       this->world->SphericalCoords()->SphericalFromLocal(position);
 #else
     const ignition::math::Pose3d pose = buoy.link->GetWorldPose().Ign();
-    const ignition::math::Vector3d position = 
+    const ignition::math::Vector3d position =
       this->world->GetSphericalCoordinates()->GlobalFromLocal(pose.Pos());
     const ignition::math::Vector3d latlon =
       this->world->GetSphericalCoordinates()->SphericalFromLocal(position);
--- a/vrx_gazebo/src/wildlife_scoring_plugin.cc
+++ b/vrx_gazebo/src/wildlife_scoring_plugin.cc
@@ -461,12 +461,14 @@ void WildlifeScoringPlugin::PublishAnimalLocations()
     // Conversion from Gazebo Cartesian coordinates to spherical.
 #if GAZEBO_MAJOR_VERSION >= 8
     const ignition::math::Pose3d pose = buoy.link->WorldPose();
-    const ignition::math::Vector3d position = this->world->SphericalCoords()->GlobalFromLocal(pose.Pos());
+    const ignition::math::Vector3d position = 
+      this->world->SphericalCoords()->GlobalFromLocal(pose.Pos());
     const ignition::math::Vector3d latlon =
       this->world->SphericalCoords()->SphericalFromLocal(position);
 #else
     const ignition::math::Pose3d pose = buoy.link->GetWorldPose().Ign();
-    const ignition::math::Vector3d position = this->world->GetSphericalCoordinates()->GlobalFromLocal(pose.Pos());
+    const ignition::math::Vector3d position = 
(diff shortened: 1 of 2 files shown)


vrx_gazebo/src/wildlife_scoring_plugin.cc
```diff
@@ -461,12 +461,14 @@ void WildlifeScoringPlugin::PublishAnimalLocations()
     // Conversion from Gazebo Cartesian coordinates to spherical.
 #if GAZEBO_MAJOR_VERSION >= 8
     const ignition::math::Pose3d pose = buoy.link->WorldPose();
+    const ignition::math::Vector3d position = this->world->SphericalCoords()->GlobalFromLocal(pose.Pos());
```
caguero, line 464: Could you split this line into two lines? The linter complains because it's > 80 chars.

vrx_gazebo/src/wildlife_scoring_plugin.cc
```diff
@@ -461,12 +461,14 @@ void WildlifeScoringPlugin::PublishAnimalLocations()
...
     const ignition::math::Vector3d latlon =
-      this->world->SphericalCoords()->SphericalFromLocal(pose.Pos());
+      this->world->SphericalCoords()->SphericalFromLocal(position);
 #else
     const ignition::math::Pose3d pose = buoy.link->GetWorldPose().Ign();
+    const ignition::math::Vector3d position = this->world->GetSphericalCoordinates()->GlobalFromLocal(pose.Pos());
```
caguero, line 469: Could you split this line into two lines? The linter complains because it's > 80 chars.

caguero (changes requested):
Thanks for the contribution. I just added two minor comments because the linter complains.

vrx_gazebo/src/wildlife_scoring_plugin.cc
alexperez33: No problem!

vrx_gazebo/src/wildlife_scoring_plugin.cc
alexperez33: No problem!

j-herman:
@caguero I've been digging into the SphericalCoordinates source code this afternoon to try to figure out this issue.  

Our world frame (which should be CoordinateType LOCAL) is ENU, and thus coincident with GLOBAL.  I validated this with a call to HeadingOffset() that returned 0, just to be sure.  So, in theory, the call to GlobalFromLocal() shouldn't do anything at all.  In fact, though, conversions from LOCAL to anything else swap the signs on the X and Y inputs.  This happens in both the PositionTransform() and VelocityTransform() methods.  
I also validated that the sign swap is what's happening in our wildlife worlds - can provide code snippet if desired.

Why this fix appears to work: in the call to GlobalFromLocal(), the signs are swapped.  Then in the call to SphericalFromLocal(), they're swapped back.  

The temporary fix for us should be to do the transform from GLOBAL to SPHERICAL via PositionTransform().  

I am going to do more math and see if I can figure out why the functions were written this way and what should really be happening.

j-herman:
Just as a clear demonstration of the error, here's some output from the default wayfinding world.  We enter the waypoints as lat/long, and convert via SphericalToLocal(), which works correctly and places the marker where we expect.  I added the reverse conversion to show what the LocalToSpherical() function returns - you can see the differences.  SphericalToLocal assumes ENU, while LocalToSpherical is reversed in X,Y.

[Msg] Waypoint, Spherical: Lat = -33.7227 Lon = 150.674
[Msg] Waypoint, Local: X = 525.794 Y = -171.5 Yaw = 1.21756
[Msg] Waypoint, converted from local back to spherical: -33.7258 150.685 1.24152

[Msg] Waypoint, Spherical: Lat = -33.7221 Lon = 150.674
[Msg] Waypoint, Local: X = 550.984 Y = -237.315 Yaw = 1
[Msg] Waypoint, converted from local back to spherical: -33.7264 150.686 1.02821

[Msg] Waypoint, Spherical: Lat = -33.7226 Lon = 150.675
[Msg] Waypoint, Local: X = 434.679 Y = -179.863 Yaw = 1
[Msg] Waypoint, converted from local back to spherical: -33.7258 150.684 1.01734

j-herman:
This is an issue in the Stationkeeping scoring plugin as well, but only in the case where the goal pose is entered in cartesian coords rather than lat/long, which is probably why we haven't seen it in practice.
For style discussion before making any recommendations on this pull request, I've got proposed changes in a new branch: jessica/spherical_fix

caguero:
@alexperez33 , @j-herman, I'm going to close this pull request in favor of #428 , as I think it properly handles this issue. Thanks a lot for the useful discussion here.

PR #428, opened 2022-02-07
Fix spherical coordinates FromLocal calls

caguero:
See issue #420 and pull request #421 where there's a detailed conversation.

As described in [this pull request](https://github.com/ignitionrobotics/ign-math/pull/235), the `*FromLocal*()` functions contain a bug that has been fixed in Ignition Math. This patch instantiates a new ignition::math::SphericalCoordinates member variable in the base scoring plugin to be used for any derived scoring plugins.

Note that even in the updated version we still can't use `*FromLocal*()` directly, as it contains the version with the issue to preserve old behavior. In order to use the corrected version we need to call `PositionTransform()` directly.

### DIFF
--- a/vrx_gazebo/config/wamv_config/component_compliance/numeric.yaml
+++ b/vrx_gazebo/config/wamv_config/component_compliance/numeric.yaml
@@ -61,3 +61,8 @@ wamv_ball_shooter:
         pitch
         yaw
         name
+wamv_pinger:
+    num: 1
+    allowed_params:
+        position
+        name
--- a/vrx_gazebo/include/vrx_gazebo/scoring_plugin.hh
+++ b/vrx_gazebo/include/vrx_gazebo/scoring_plugin.hh
@@ -26,6 +26,7 @@
 #include <gazebo/common/Events.hh>
 #include <gazebo/common/Plugin.hh>
 #include <gazebo/common/Time.hh>
+#include <ignition/math/SphericalCoordinates.hh>
 #include <gazebo/physics/World.hh>
 #include <sdf/sdf.hh>
 #include <gazebo/transport/transport.hh>
@@ -232,6 +233,9 @@ class ScoringPlugin : public gazebo::WorldPlugin
 
   /// \brief Silent mode enabled?
   protected: bool silent = false;
+
+  /// \brief Spherical coordinates conversions.
+  protected: ignition::math::SphericalCoordinates sc;
 
   /// \brief gazebo node pointer
   private: gazebo::transport::NodePtr gzNode;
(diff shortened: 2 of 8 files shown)


caguero:
@M1chaelM , `test-gazebo11-noetic` fails because the Docker image `npslearninglab/watery_robots:noetic_current` probably keeps an old Ignition Math version. Do you think you could generate a new image with the latest Ignition Math version (`6.10`)?

j-herman:
@caguero I checked out the code on my local installation and all is working correctly in both the stationkeeping and wildlife transforms for noetic - just waiting on the docker fix.  I am having trouble getting anything to run in the melodic docker image so will have to retest later.
I think we could use PositionTransform directly in the earlier versions as well, with 'GLOBAL' instead of 'LOCAL2', to keep the syntax consistent, but this should work so whatever you prefer works for me.

caguero:
> @caguero I checked out the code on my local installation and all is working correctly in both the stationkeeping and wildlife transforms for noetic - just waiting on the docker fix. I am having trouble getting anything to run in the melodic docker image so will have to retest later. I think we could use PositionTransform directly in the earlier versions as well, with 'GLOBAL' instead of 'LOCAL2', to keep the syntax consistent, but this should work so whatever you prefer works for me.

I just updated the code. Is that what you're thinking?

j-herman:
Yes, that's what I was thinking, or we can even simplify further (stationkeeping example):
``` 
   ignition::math::Vector3d cartVec(this->goalX, this->goalY, xyz.Z());
   ignition::math::Vector3d scVec;

#if GAZEBO_MAJOR_VERSION >= 11
    auto in = ignition::math::SphericalCoordinates::CoordinateType::LOCAL2;
    auto out = ignition::math::SphericalCoordinates::CoordinateType::SPHERICAL;
#else
    auto in = gazebo::common::SphericalCoordinates::CoordinateType::GLOBAL;
    auto out = gazebo::common::SphericalCoordinates::CoordinateType::SPHERICAL;
#endif

  scVec = this->sc.PositionTransform(cartVec, in, out);
  scVec.X(IGN_RTOD(scVec.X()));
  scVec.Y(IGN_RTOD(scVec.Y()));
```
Maybe it would be easiest to use GLOBAL for all versions for now, until Michael has time to update the docker image?

caguero:
> 0060993

I like your last suggestion. How about this? [0060993](https://github.com/osrf/vrx/pull/428/commits/00609931b4790f67ecde7ba676d1927c583b2f1a)

j-herman (approved):
Checked both wildlife and stationkeeping tasks - spherical coordinates are correct in both.  Good to merge!

PR #429, opened 2022-02-11
Jessica/phase2 updates

j-herman:
After reviewing the code changes needed I made the changes to keep the <radius> tag.  I will change the values to reflect this in the existing worlds, so the animal buoy behavior doesn't change from what was reviewed.


caguero (changes requested):
The changes look good to me. I suggest one more change in the line section. Something like this:
```
else if (_sdf->HasElement("line"))
  {
    auto lineElem = _sdf->GetElement("line");
    // Parse the required <direction> field.
    if (!lineElem->HasElement("direction"))
    {
      ignerr << "No <line><direction> specified" << std::endl;
      return;
    }
    ignition::math::Angle direction =
      lineElem->Get<ignition::math::Angle>("direction");

    // Parse the required <length> field.
    if (!lineElem->HasElement("length"))
    {
      ignerr << "No <line><length> specified" << std::endl;
      return;
    }
    auto length = lineElem->Get<double>("length");

    // Create a relative vector in the direction of "direction" and of
    // length "length".
    ignition::math::Vector3d lineVec(
      length * cos(direction.Radian()),
      length * sin(direction.Radian()), 0);
    ignition::math::Vector2d position(this->modelPose.Pos().X(),
                                      this->modelPose.Pos().Y());
    // Add the initial model position and calculated endpoint as waypoints.
    this->localWaypoints.push_back(position);
    ignition::math::Vector3d p = this->modelPose.CoordPositionAdd(lineVec);
    ignition::math::Vector2d p2D = {p.X(), p.Y()};
    this->localWaypoints.push_back(p2D);
    igndbg << "Entered line waypoints " << position << ", " << p2D << std::endl;
```

I think that the current code has an issue with the line being relative to the robot position. I think this version fixes it. I also suggest to use radians to make the angles more consistent with other places in Gazebo.

j-herman:
@caguero Going through this it looks like there are three things you are suggesting for changes:
1.  Remove the default values for direction and distance, and make those parameters required (should we change this in the circle mode, too?)
2. Specify the direction of motion in Gazebo coordinates/ conventions when initializing the follow plugin (ENU, radians, etc.)
3. The main issue with the direction of motion is that the documentation is wrong - it's already setting the direction of motion from the initial model position in ENU coordinates, so a <direction> tag value of 0 will have the model moving to the east, 090 will move the model towards the north, etc.  We can use the CoordPosAdd functionality if you think it makes the math easier to follow but I've confirmed both sets of code are producing the same output.  Could you describe the behavior that isn't right, as far as the line of motion being relative to the model's position?

caguero:
> @caguero Going through this it looks like there are three things you are suggesting for changes:
> 
> 1. Remove the default values for direction and distance, and make those parameters required (should we change this in the circle mode, too?)
> 2. Specify the direction of motion in Gazebo coordinates/ conventions when initializing the follow plugin (ENU, radians, etc.)
> 3. The main issue with the direction of motion is that the documentation is wrong - it's already setting the direction of motion from the initial model position in ENU coordinates, so a  tag value of 0 will have the model moving to the east, 090 will move the model towards the north, etc.  We can use the CoordPosAdd functionality if you think it makes the math easier to follow but I've confirmed both sets of code are producing the same output.  Could you describe the behavior that isn't right, as far as the line of motion being relative to the model's position?

(1) is a suggestion to make the behavior more predictable where the user needs to set the parameters. About (2) and (3) my intention is to make the angle relative to the initial position of the model, not an angle in the world frame. E.g.: If the angle is set to `0`, the robot will always continue straight until the right distance is covered. This should be the behavior with any initial yaw angle of the model. I think the original code doesn't behave like that.

j-herman:
> 
> (1) is a suggestion to make the behavior more predictable where the user needs to set the parameters. About (2) and (3) my intention is to make the angle relative to the initial position of the model, not an angle in the world frame. E.g.: If the angle is set to `0`, the robot will always continue straight until the right distance is covered. This should be the behavior with any initial yaw angle of the model. I think the original code doesn't behave like that.

Ah, now I get it!  That makes sense.  I'll make the changes.

j-herman:
@caguero The latest commit should address your comments.  I checked that the relative heading is working as we discussed: with the <direction> tag set to pi/2, the animal makes a rough 90 degree left turn from the heading set in its initial pose.  I've also updated the <circle> code to make the <radius> tag required, and changed the comments to make it more clear that the line behavior is steering to waypoints rather than following a particular path.
I am going to wait to update the yamls for world generation in case there's anything else I'm missing here, but that will be quick and easy once this is through.  Thanks for the review!

caguero (approved):
It looks good to me!

Issue #431, opened 2022-02-22
Some examples use wind parameters outside the specified envelope


M1chaelM:
Our documentation specifies that the value of `wind_variance_time` should be between 2 and 20. However, we have some practice example worlds in which it is set to 1. This could be misleading, so we should update them so they are at least 2.

M1chaelM:
This is partially addressed by PR #432, but needs to also be addressed upstream in the yaml files in the `vrx-docker` repository.

j-herman:
@M1chaelM I have made updates to all the yamls in the phase2 repo - I can make the changes to these files now, or once we add the phase2 worlds as additional practice.

M1chaelM:
@j-herman I think the `vrx-docker` side of this was addressed in https://github.com/osrf/vrx-docker/pull/44 so we may be able to close this now. Is there anything else to do?

PR #432, opened 2022-02-22
updating all var_wind_time 1 to 2 to keep within the envelope

M1chaelM:
Partially addresses issue #431 by modifying all values for the var_wind_time parameter so they are within the envelop. (Note that all but one of them were set to 1, so they are now all set to 2. Apparently we aren't relying on this parameter very much.)

Since the change is very minor, I don't think it's necessary to test every example world. To review, please look at the code changes to double check that nothing was changed except to increase `var_wind_time_constants` from 1 to 2, and that this was done for all 18 example files in `vrx_gazebo/worlds/2022_practice` (one of these was already 2, so 17 are changed).

You might also run one of the worlds just to make sure:

```
WORLD=stationkeeping0.world
roslaunch vrx_gazebo vrx.launch verbose:=true \
	  paused:=false \
	  wamv_locked:=true \
	  world:=${HOME}/vrx_ws/src/vrx/vrx_gazebo/worlds/2022_practice/${WORLD}
```

### DIFF
--- a/vrx_gazebo/worlds/2022_practice/gymkhana0.world
+++ b/vrx_gazebo/worlds/2022_practice/gymkhana0.world
@@ -160,7 +160,7 @@
       <!-- in degrees -->
       <wind_mean_velocity>0.0</wind_mean_velocity>
       <var_wind_gain_constants>0</var_wind_gain_constants>
-      <var_wind_time_constants>1</var_wind_time_constants>
+      <var_wind_time_constants>2</var_wind_time_constants>
       <random_seed>10</random_seed>
       <!-- set to zero/empty to randomize -->
       <update_rate>10</update_rate>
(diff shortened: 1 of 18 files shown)


j-herman:
@M1chaelM These world files are all good to go.  I saw one other file we might want to change - should we update the default value in usv_wind_plugin.xacro?  I'll drop in a new commit on this branch since I have the file up, so just reject it if you don't think that's a good addition.

caguero (approved):
Thanks, looks good!

M1chaelM:
> @M1chaelM These world files are all good to go. I saw one other file we might want to change - should we update the default value in usv_wind_plugin.xacro? I'll drop in a new commit on this branch since I have the file up, so just reject it if you don't think that's a good addition.

@j-herman Sure, I think that can't hurt.

PR #436, opened 2022-03-10
workaround for generate worlds

M1chaelM:
This change to the generate_worlds.py script eliminates the need to call `include_all_macros.xacro` by writing the include files directly into the generated `xacro`. This workaround allows us to continue to generate worlds until the scope issue for includes in `xacro` is resolved.


caguero:
Clossing this one for now, as we'll be using `main`.

M1chaelM:
This may still be needed to get the `generate_worlds` script to work correctly in the `gazebo_classic` branch. I need to look back, but think I was actually running off this branch when we evaluated the submissions for vrx 2022.

PR #439, opened 2022-03-15
Removed set pinger position node from vrx.launch as well

alexperez33:
@caguero I was able to replicate your fix with success. I noticed that you removed the set_pinger_position node from gymkhana.launch. Would it make sense to remove it from vrx.launch as well (as shown in this PR)? Personally, I have my code set up such that vrx.launch is always called and the world file we want is passed in as a separate arg. I am not sure if anyone else does it like this, but if they did, they would still run into the pinger issue.

### DIFF
--- a/vrx_gazebo/launch/vrx.launch
+++ b/vrx_gazebo/launch/vrx.launch
@@ -68,9 +68,6 @@
               -urdf -param $(arg namespace)/robot_description -model wamv"/>
 
   <!-- Set the pinger location -->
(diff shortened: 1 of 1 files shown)


caguero (approved):
Thanks for the patch @alexperez33 !. Using `vrx.launch` and setting the `world` parameter is correct. In fact this is how we do it in `vrx-docker` to launch trials.

The only issue here is that if you have a world without a gymkhana scoring plugin, nobody will set the position of the pinger. For example if you launch:
```
```

The pinger location will default to origin unless you manually call `set_pinger_position.py` . I think this is not very unreasonable so I'm approving.

Issue #445, opened 2022-04-01
Unable to release joints in "ready state" because they do not exist


alexperez33:
When running any of the tasks, I noticed that I would always get the error ```Unable to release [wamv_external_pivot_joint]``` and ```Unable to release [wamv_external_riser]```. I also noticed that my urdf of the boat does not have these joints, which is generated using the generate_wamv.launch script. This seems to cause the boat to not be locked in place during the initial state of a task. This is especially problematic during tasks with big waves and wind because we don't move the boat until the ready state.

M1chaelM:
@alexperez33 Could you please post the command you are using to run the tasks when you see this error?

alexperez33:
@M1chaelM The error occurs whenever I launch ```vrx.launch``` using the urdf that was generated with the ```generate_wamv.launch``` script. This is due to the fact that the urdf that was generated with the ```generate_wamv.launch``` script does not contain the two joints specified above.

The script ```generate_wamv.launch``` does not allow us to send a parameter into the ```wamv_gazebo.urdf.xacro``` to override the default "locked" parameter (which is false). This "locked" parameter is in a xacro conditional which adds the above two joints.

All this to say, because my urdf does not contain these two joints, I receive the above errors and the boat does not stay locked during the initial state.

I run the generate_wamv.launch file as such:
```roslaunch vrx_gazebo generate_wamv.launch thruster_yaml:=<path_to_thruster_config> component_yaml:=<path_to_component_config> wamv_target:=<target_location_for_urdf_to_be_generated>```

The solution I provided in PR #446 allows a user to run the above command and add a parameter to turn the "locked" parameter true.

Another solution would be to just make the default parameter for "locked" in ```wamv_gazebo.urdf.xacro``` to true.

**Comment by M1chaelM** [comment]
@alexperez33 Thanks for these details, and the PR. I'm going to look into this today.

I believe the error you are describing should be produced by a scoring plugin if it attempts to release a wamv that is not locked. Launching `vrx.launch` by itself should not cause any scoring plugin to run because there is no task associated with it by default. Are you running something like:
```
roslaunch vrx_gazebo vrx.launch world:=`pwd`/stationkeeping0.world urdf:=$WAMV
```

alexperez33:
I apologize, I forgot to mention that I am passing in a world as an argument. The error is produced by scoring_plugin.cc in the ```ReleaseVehicle``` function. It causes an error because it cannot find the joints.

M1chaelM:
[...]

PR #446, opened 2022-04-01
Fixed the generate urdf scripts such that we can pass in "locked" parameter.

alexperez33:
One potential solution to issue #445 is described by this pull request. I don't know why the default parameter for the arg "locked" is false in ```wamv_gazebo.urdf.xacro```, so I left it false. This causes the xacro related to the wamv_external_pivot_joint and wamv_external_riser to be skipped. Because of this anyone who generated urdf using the generate_wamv.launch script would not have their boat locked in the initial state of a task.

Instead of changing the default parameter to true in the ```wamv_gazebo.urdf.xacro```, I added a parameter within ```generate_wamv.launch``` that can be passed in to override the default "locked" parameter of the ```wamv_gazebo.urdf.xacro``` if the user decides to do so. This fixed the problem for me.


M1chaelM:
### Testing:
1. Generate wamv's with and without the lock joints.
  ```
  cd ~/vrx_ws/src/vrx-docker
  roslaunch vrx_gazebo generate_wamv.launch thruster_yaml:=`pwd`/team_config/example_team/thruster_config.yaml component_yaml:=`pwd`/team_config/example_team/component_config.yaml wamv_target:=`pwd`/locked_wamv.urdf wamv_locked:=True
  roslaunch vrx_gazebo generate_wamv.launch thruster_yaml:=`pwd`/team_config/example_team/thruster_config.yaml component_yaml:=`pwd`/team_config/example_team/component_config.yaml wamv_target:=`pwd`/unlockable_wamv.urdf wamv_locked:=False
  ```
2. Check for the joints by running:
  ```
  grep wamv_external locked_wamv.urdf
  grep wamv_external unlockable_wamv.urdf
  ```
  The output for the first grep command is:
  ```
   <link name="wamv_external_link"/>
    <joint name="wamv_external_pivot_joint" type="universal">
      <child>wamv_external_link</child>
    <joint name="wamv_external_riser" type="prismatic">
      <child>wamv_external_link</child>
  ```
  The second produces no output.
3. Verify that the wamv is locked as expected:
  ```
  roslaunch vrx_gazebo vrx.launch world:=`pwd`/generated/task_generated/stationkeeping/worlds/stationkeeping0.world urdf:=`pwd`/locked_wamv.urdf verbose:=true
  ```
  We no longer see the error:
  ```
  [Err] [scoring_plugin.cc:248] Unable to release [wamv_external_pivot_joint]
  [Err] [scoring_plugin.cc:248] Unable to release [wamv_external_riser]
  ```

### A few other notes for our future selves:
* When we are launching a world while overriding the `urdf` argument, the value of the `wamv_locked` argument has no effect. This only has an impact when we're using `wamv_gazebo.urdf.xacro` to generate the `urdf` on the fly at launch.
* Conversely, a `urdf` that describes a wamv that was generated with `generate_wamv.launch` using this `wamv_locked` argument can only be used in conjunction with a scoring plugin. Without the scoring plugin to release it, it will remain locked.
* Overall this is not a great state of affairs. Ideally the lock should be generated and destroyed by the scoring plugin.

M1chaelM (approved):
Thank you for this contribution! Very much appreciated. Please merge as soon as you are able and we'll resolve issue #445

alexperez33:
I do not think I have the permissions to merge, but no problem! @M1chaelM

PR #449, opened 2022-04-06
Finish Gymkhana when failing crossing the navigation channel

caguero:
There is an issue with our current combination of navigation and gymkhana scoring plugins. If the navigation scoring plugin decides to terminate (because the WAM-V performs an illegal maneuver such as crossing a gate backwards or out of order), the gymkhana scoring plugin didn't know about it and continued enabling the scoring as if the team has crossed the channel.

This PR fixes this issue by checking the actual score coming from the navigation channel task message. If we detect that the score is the `timeout_score`, then, the gymkhana task shouldn't continue.

How to test it?

1. Run any gymkhana task (**change the path to the world file**):

```
roslaunch vrx_gazebo vrx.launch world:=/home/caguero/vrx2022_phase2_ws/src/vrx-docker/generated/task_generated/gymkhana/worlds/gymkhana0.world verbose:=true wamv_locked:=true
```

2. Run the joystick controller:
```
```

Teleoperate the WAM-V and make two illegal movements. Try crossing an uncrossed gate backwards and skipping a gate. The gymkhana runs should terminate. The score should be MAX.

Additionally, complete the navigation channel correctly and verify that the task continues. The score should drop substantially when crossing the channel from MAX to a few meters (70-100).

### DIFF
--- a/vrx_gazebo/src/gymkhana_scoring_plugin.cc
+++ b/vrx_gazebo/src/gymkhana_scoring_plugin.cc
@@ -123,9 +123,12 @@ void GymkhanaScoringPlugin::ChannelCallback(
     // Determine whether channel has been crossed before timeout
     if (!this->channelCrossed)
     {
-      if (msg->state == "finished" && !msg->timed_out)
+      if (msg->state == "finished")
       {
-        this->channelCrossed = true;
+        if (msg->score == this->ScoringPlugin::GetTimeoutScore())
+          this->Finish();
+        else
+          this->channelCrossed = true;
       }
     }
   }
--- a/vrx_gazebo/include/vrx_gazebo/scoring_plugin.hh
+++ b/vrx_gazebo/include/vrx_gazebo/scoring_plugin.hh
@@ -331,7 +331,7 @@ class ScoringPlugin : public gazebo::WorldPlugin
   private: ros::Publisher contactPub;
 
   /// \brief Score in case of timeout - added for Navigation task
-  private: double timeoutScore = -1.0;
+  private: double timeoutScore = 200.0;
 
   /// \brief Whether to shut down after last gate is crossed.
   private: bool perPluginExitOnCompletion = true;
--- a/vrx_gazebo/src/navigation_scoring_plugin.cc
+++ b/vrx_gazebo/src/navigation_scoring_plugin.cc
@@ -140,12 +140,6 @@ void NavigationScoringPlugin::Load(gazebo::physics::WorldPtr _world,
   // Save number of gates
   this->numGates = this->gates.size();
 
-  // Set default score in case of timeout.
-  double timeoutScore = 2.0 * this->GetRunningStateDuration() /
-                        static_cast<double>(this->numGates);
-  gzmsg << "Setting timeoutScore = " << timeoutScore << std::endl;
-  this->ScoringPlugin::SetTimeoutScore(timeoutScore);
-
   gzmsg << "Task [" << this->TaskName() << "]" << std::endl;
 
   this->updateConnection = gazebo::event::Events::ConnectWorldUpdateBegin(


M1chaelM:
I'm seeing that the trial now terminates instantly when there is **one** illegal movement:
```
[Msg] Transited the gate in the wrong direction. Gate invalidated!
[Err] [scoring_plugin.cc:502] VRX_EXIT_ON_COMPLETION and <per_plugin_exit_on_completion> both not set, will not shutdown on ScoringPlugin::Exit()
[ERROR] [1649280793.997056573, 46.088000000]: VRX_EXIT_ON_COMPLETION and <per_plugin_exit_on_completion> both not set, will not shutdown on ScoringPlugin::Exit()
```
The score reported in `/vrx/task/info` starts at 0 but then jumps to:
```
score: 1.7976931348623157e+308
```
as soon as the state becomes `onrunning`. Is that the max?

M1chaelM:
I'm looking back at our documentation for this task and I think it might need some clarification. We say that if you don't cross the channel, your score defaults to the "maximum error" but I don't see anywhere where we say what the maximum error is. Hopefully it's not `1.7976931348623157e+308` since that would be tough to recover from :laughing: . I see in the code it looks like we're setting it to 200. That seems reasonable as long as no one can score worse than that (for example by driving away from the buoy), but I do think we should write it down somewhere.

I'm also not sure about limiting teams to two illegal movements. If we have that in the rules somewhere it's no problem but I'm not seeing it. The way the rules are written it seems like it would be a less radical change to go with one of the following guidelines:
* You have completed the navigation channel whenever you cross all gates in the correct order, regardless of how many illegal moves you make in between, OR
* You are not allowed to make more than one illegal move, and If you do you have to start over from the first gate.

What do you think? Sorry if I'm missing this in the documentation somewhere--please let me know if that's the case.

caguero:
The navigation channel sub rules are inherited from VRX'19, where teams were not allowed to miss a single gate or crossed incorrectly. The rules for this year are still capturing that essence. The note about getting the maximum error means that your run should terminate with a score of `1.7976931348623157e+308`. I feel that we shouldn't change it at this point unless we see a bit problem with it.

caguero:
> I'm looking back at our documentation for this task and I think it might need some clarification. We say that if you don't cross the channel, your score defaults to the "maximum error" but I don't see anywhere where we say what the maximum error is. Hopefully it's not `1.7976931348623157e+308` since that would be tough to recover from laughing . I see in the code it looks like we're setting it to 200. That seems reasonable as long as no one can score worse than that (for example by driving away from the buoy), but I do think we should write it down somewhere.
> 
> I'm also not sure about limiting teams to two illegal movements. If we have that in the rules somewhere it's no problem but I'm not seeing it. The way the rules are written it seems like it would be a less radical change to go with one of the following guidelines:
> 
> * You have completed the navigation channel whenever you cross all gates in the correct order, regardless of how many illegal moves you make in between, OR
> * You are not allowed to make more than one illegal move, and If you do you have to start over from the first gate.
> 
> What do you think? Sorry if I'm missing this in the documentation somewhere--please let me know if that's the case.

The navigation channel rules are inherited from VRX'19, where teams were not allowed to miss a single gate or cross it incorrectly. The rules for this year are still capturing that essence. The note about getting the maximum error means that your run should terminate with a score of `1.7976931348623157e+308`. I feel that we shouldn't change it at this point unless we see a bit problem with it.

caguero:
> I'm seeing that the trial now terminates instantly when there is **one** illegal movement:
> 
> ```
> [Msg] Transited the gate in the wrong direction. Gate invalidated!
> [Err] [scoring_plugin.cc:502] VRX_EXIT_ON_COMPLETION and <per_plugin_exit_on_completion> both not set, will not shutdown on ScoringPlugin::Exit()
> [ERROR] [1649280793.997056573, 46.088000000]: VRX_EXIT_ON_COMPLETION and <per_plugin_exit_on_completion> both not set, will not shutdown on ScoringPlugin::Exit()
> ```
> 
> The score reported in `/vrx/task/info` starts at 0 but then jumps to:
> 
> ```
> score: 1.7976931348623157e+308
> ```
> 
> as soon as the state becomes `onrunning`. Is that the max?

Yes, it is.

M1chaelM:
> The navigation channel rules are inherited from VRX'19, where teams were not allowed to miss a single gate or cross it incorrectly. The rules for this year are still capturing that essence. The note about getting the maximum error means that your run should terminate with a score of `1.7976931348623157e+308`. I feel that we shouldn't change it at this point unless we see a bit problem with it.

I think there must be some confusion here. In 2019, the score for the navigation task was simply the amount of time it took the teams to complete the course. If they did not complete it, the score was set to the maximum run time. We gave a formula to calculate maximum run time, and it was:

> The maximum run time (T) for each run is T=D/V, where V is a nominal speed of 0.25 m/s and D is the total piecewise linear distance between gates.

It would be very strange for maximum run time to be set to the largest representable number. Further, since the task score is the mean of all run scores, failure to finish even one channel would be devastating.

In the gymkhana task, the score is measured in a different unit from the navigation channel: it's an error distance in meters rather than a completion time in seconds. To keep with the spirit of the 2019 rules, I think it would make much more sense to set the maximum value equal to the initial distance between the WAMV at its starting position and the pinger.

vrx_gazebo/src/gymkhana_scoring_plugin.cc
```diff
@@ -123,9 +123,12 @@ void GymkhanaScoringPlugin::ChannelCallback(
     // Determine whether channel has been crossed before timeout
     if (!this->channelCrossed)
     {
-      if (msg->state == "finished" && !msg->timed_out)
```
M1chaelM, line 126: Removing this condition causes the task to exit immediately when a gate is crossed incorrectly. The  `navigation_scoring_plugin` reports a finished state, and the score is equal to the timeout score, so it finishes. Is this intended? I'm pointing this out because it seems like your testing instructions suggest that I should be able to make at least two illegal moves.

M1chaelM:
The score looks good now, but when an illegal move is committed, the task time looks like it switches to nanoseconds before exiting:

```
name: "gymkhana"
state: "finished"
ready_time: 
  secs: 10
  nsecs:         0
running_time: 
  secs: 20
  nsecs:         0
elapsed_time: 
  secs: 17
  nsecs: 956000000
remaining_time: 
  secs: 282
  nsecs:  44000000
timed_out: False
num_collisions: 0
score: 200.0
```

M1chaelM (approved):
See the minor note about the strange switch to nanoseconds. Otherwise this all works as advertised.

caguero:
> The score looks good now, but when an illegal move is committed, the task time looks like it switches to nanoseconds before exiting:
> 
> ```
> name: "gymkhana"
> state: "finished"
> ready_time: 
>   secs: 10
>   nsecs:         0
> running_time: 
>   secs: 20
>   nsecs:         0
> elapsed_time: 
>   secs: 17
>   nsecs: 956000000
> remaining_time: 
>   secs: 282
>   nsecs:  44000000
> timed_out: False
> num_collisions: 0
> score: 200.0
> ```
> 
> Looking closely, this also happens when the regular timeout occurs but it is less noticeable when the value is 0. It's strange but seems like an unrelated issue.

Could you please clarify what do you mean by the task time looks like it switches to nanoseconds?

M1chaelM:
> Could you please clarify what do you mean by the task time looks like it switches to nanoseconds?

Looking more closely, maybe it's functioning as intended. I was confused because the `nsecs` fields are always 0 except when the task terminates do to an illegal move, at which point they are suddenly populated. This is probably because the `/vrx/task/info` message only updates at a frequency of 1Hz.

caguero:
> > Could you please clarify what do you mean by the task time looks like it switches to nanoseconds?
> 
> Looking more closely, maybe it's functioning as intended. I was confused because the `nsecs` fields are always 0 except when the task terminates do to an illegal move, at which point they are suddenly populated. This is probably because the `/vrx/task/info` message only updates at a frequency of 1Hz.

Exactly, I think this is expected.

Issue #451, opened 2022-04-09
Placard subscriber seems to sometimes instantiate after new dock symbols are published


alexperez33:
This issue pertains to the Scan Dock and Deliver task.

I was not able to replicate this issue using only the vrx related launch files. However, when I launch my code alongside the vrx code, I noticed that sometimes the publisher within ```scan_dock_scoring_plugin.cc``` within the AnnounceSymbol function publishes the dock symbols before the subscriber within ```placard_plugin.cc``` is even instantiated. This is not always the case, but when it is the case, the dock simply ends up showing the default "blue_circle" symbols because the subscriber never receives the actual symbols published from  ```scan_dock_scoring_plugin.cc```. (The dock symbols are only published once).

I am not sure exactly why it takes so long for the subscriber in ```placard_plugin.cc``` to instantiate. Perhaps there is a lot of other code that is starting up simultaneously which could cause this delay. Regardless, I believe the solution for this would be to make the gazebo subscriber within ```placard_plugin.cc``` allow latching. This way once even if the msgs are published before the subscriber is instantiated, the subscriber can still just get the last published message.

M1chaelM:
Resolved with #452

PR #452, opened 2022-04-09
Made placard symbol subscriber enable latching

alexperez33:
The solution that solved issue #451 for me was to make the gazebo subscriber within placard_plugin.cc allow latching. This way once even if the msgs are published before the subscriber is instantiated, the subscriber can still just get the last published message.

@M1chaelM, I apologize that I cannot think of a way to for you to replicate the issue without running my code, however, I think it makes sense as well to make this subscriber enable latching since the placard symbols are only ever published once. Please let me know if you are interested in seeing the behavior I am describing in #451 more in depth. You should also see that this change does not break anything, at least to my knowledge.

Update: I recently encountered the same issue for the STC buoy. It seemed that the subscriber was too late to instantiate, and it did not receive the published msg containing the lights information from the ```scan_dock_scoring_plugin.cc```. Therefore, the STC buoy was always black. The fix was to make this STC code related subscriber also latching.

This is the documentation I referenced: 
https://osrf-distributions.s3.amazonaws.com/gazebo/api/1.4.0/classgazebo_1_1transport_1_1Node.html#a92eab7694fab5eaf46c8ea046212e662

### DIFF
--- a/vrx_gazebo/src/light_buoy_plugin.cc
+++ b/vrx_gazebo/src/light_buoy_plugin.cc
@@ -136,7 +136,7 @@ void LightBuoyPlugin::Load(gazebo::rendering::VisualPtr _parent,
 
   gzNode->Init();
   this->colorSub = this->gzNode->Subscribe
-    (this->gzColorsTopic, &LightBuoyPlugin::ChangePatternTo, this);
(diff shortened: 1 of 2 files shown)


caguero (approved):
This looks reasonable to me.

PR #467, opened 2022-07-04
Removed 'loop_forever' from perception scoring and perception task world

### LINKED ISSUE #457: Scoring for Perception Task
Hello,

When my team runs the perception task we are correctly identifying which buoy is where (left to right) but with some localization inaccuracy. We expect to incur 2m of error per buoy, and in fact the debug message that the perception_task publishes to the command window shows that.
I believe based on how the scoring reads, that if we get an error of 2 for each buoy, our overall score should then be 2. However, the score slowly builds up over the course of the run to be somewhere around 2.888. I know that improving our localization accuracy will improve this score, I'm just a bit confused as to how this math breaks down for the scoring. I've attached an image below showing the debug message I'm talking about. Please let me know if we're misinterpreting the scoring calculation. Thanks.
image

rmvanarse:
This PR removes the use of the `loop_forever` parameter in the scoring plugin for the perception task to resolve #457  (the course needs to be fully specified now). It also removes the line setting it to true in `perception_task.world.xacro`.


caguero (changes requested):
Could you also remove the documentation associated to this parameter in `perception_scoring_plugin.hh`?

rmvanarse:
Removed the declarations and documentation for `loop_forever` from the header

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,20 @@
+{
```
caguero, line 1: Could you remove your `.vscode` files?

caguero (changes requested):

caguero (approved):

PR #470, opened 2022-07-12
Add components from the VORC project.

bsb808:
### Description

This PR moves the VORC project functionality to the VRX project.  This will allow us to maintain another use-case for the project in the same repository and simplify regression testing.

The PR includes two ROS packages copied directly from VORC.

1. `cora_description`: mesh, URDF, etc. for the CoRA USV.
2. `vorc_gazebo`: harbor environment model, gymkhana scoring plugin, and other elements to run VORC scenarios.

## Documentation

Added example wiki to list of project scenarios:  https://github.com/osrf/vrx/wiki/ExampleVorc 

Will also put in PR for VORC project to add note to README pointing new work to VRX.

## How to Test

Verify functionality with this tutorial: https://github.com/osrf/vorc/wiki/Marina-Basic


M1chaelM:
* The CI is failing because PR #475 has not yet been reviewed / merged. Once it has been merged we can pull it into all the outstanding PRs (including this one).
* Tested in `noetic` rather than `melodic`, and with `catkin_make` rather than `colcon build`, but the tutorial runs fine.
* I added a note to the new [example tutorial](https://github.com/osrf/vrx/wiki/ExampleVorc) explaining that the tutorials on the VORC wiki expect `melodic` and `colcon`. I think the right thing to do is probably to leave these tutorials as-is for now, but eventually copy anything relevant over to the VRX Wiki and update it so it reflects our system requirements and installation instructions.

M1chaelM (changes requested):
* Although the Marina basic tutorial works, I'm noticing that this PR also adds all the tasks for the VORC competition. I may have misunderstood but I thought the intent of this PR was just to bring in the model and the marina.
* If we do want to include all the tasks and scoring plugins I wonder if we should also test them before merging this. My memory is that this is where we encountered conflicts with VRX previously (though maybe they have been resolved?).
* Also, it looks like we are duplicating the `pinger_visualization.py` and `set_pinger_position.py` files. Could we re-organize so we have only one copy of this code?

bsb808:
I cleaned up the files associated with the PR.  Should now just be the CoRA vessel description and marina model.

M1chaelM (approved):
Looks good! Thank you!

PR #473, opened 2022-07-15
Draft: Temporary skeleton code for scoring plugin

rmvanarse:
This PR contains the ported code for `ScoringPlugin.hh` and a basic skeleton for `ScoringPugin.cc.` The purpose of this is only for temporarily having something in place so that it doesn't block anyone else's work (if it depends on the scoring plugin), while I work on finalizing the implementations in ScoringPlugin.cc.

**Note:** The code in this PR follows a similar structure as `vrx_gazebo`, i.e. some messages have been implemented in ROS 2 instead of ignition transport. I am currently working on changing the implementations to `ignition::transport`. I have also ported the source code for `ScoringPlugin.cc` locally and can include that in this PR too, if required.

**Issues:**
- I had to hard-code two include_dirs in CMakeLists.txt since the compiler wasn't able to find these package paths. I have therefore kept this request as a draft right now. Can you look into what I might have to do differently in CMake so that these paths are found?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ build_*
 *.*~
 vrx_gazebo/models/dock_permutations/*
 *.pyc
+.vscode
+.vscode/*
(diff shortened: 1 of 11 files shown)


caguero:
Scoring plugin was already ported. Closing this one...

Issue #477, opened 2022-07-18
Add more features to Thruster Plugin


acxz:
This issue is a feature request to add specific features to our thruster plugin: https://github.com/osrf/vrx/blob/master/usv_gazebo_plugins/src/usv_gazebo_thrust_plugin.cc

These features are inspired by UUV's thruster plugin: https://github.com/uuvsimulator/uuv_simulator/blob/bfb40cb153684a0703173117b6bbf4258e8e71c5/uuv_gazebo_plugins/uuv_gazebo_plugins/src/ThrusterPlugin.cc

[ ] `mappingType` of `2`, `LinearInterp`. This allows users to input values from a datasheet and the thrust plugin will linearly interpret b/t them. Example user interface:
```
<mappingType>2</mappingType>
<inputValues>-1.0 -0.8 -0.6 -0.4 -0.2 0 0.2 0.4 0.6 0.8 1.0</inputValues>
<outputValues>-19.88 -16.52 -12.6 -5.6 -1.4 0.0 2.24 9.52 21.28 28.0 33.6</outputValues>
```

Example logic:
```
cmd = 0.9;
(33.6 - 28.0)/(1.0 - 0.8) * (0.9 - 0.8) + 28
```

[ ] thruster dynamics, first order model
Example user interface:
```
<dynamics>
  <type>FirstOrder</type>
  <timeConstant>0.5</timeConstant>
</dynamics>
```

[ ] minCmd and maxCmd clamping, while `maxCmd` is already implemented, it is not necessarily being used to clamp the input value.

PR #479, opened 2022-07-18
Add a minCmd field for the Thruster Plugin

acxz:
This PR attempts to chip away at #477 by adding a separate minCmd field for the thruster plugin. As of now this is primarily used in mappingType=1 for scaling the negative portion of the thrust profile. Later I'd like to add its use (along with maxCmd) as a clamp on the input commands.

P.S. ignore the c++17 commits for now


caguero:
The `jazzy` branch uses the upstream `Thruster` system from `gz-sim`. This system already implements `cmd_min` and `cmd_max`. See [here](https://github.com/gazebosim/gz-sim/blob/gz-sim9/src/systems/thruster/Thruster.hh#L85-L88).

PR #480, opened 2022-07-18
Add clamping

acxz:
This PR adds input and output clamping as specified in #477 

supersedes and closes #479


caguero:
The `jazzy` branch uses the `Thruster` system from `gz-sim`. This system already clamps the value to apply. See [here](https://github.com/gazebosim/gz-sim/blob/gz-sim9/src/systems/thruster/Thruster.cc#L520).

PR #481, opened 2022-07-19
Add linear interp

acxz:
related issue: #477 

From commit message:

> this mappingType allows users to specify a mapping of thrust commands and thrust forces as seen typically in thrust profile datasheets; linear interpolation using the two nearest points in the mapping is done to compute the actual thrust force from a thrust command

Again if you could ignore the previous commits for now and just review based on the latest commit prevalent that would be appreciated. I'll rebase this branch based on the input of the other PRs I have made. Sorry about that, its just easier for me to develop. Thank you.

### DIFF
--- a/usv_gazebo_plugins/CMakeLists.txt
+++ b/usv_gazebo_plugins/CMakeLists.txt
@@ -19,8 +19,8 @@ catkin_package(
   CATKIN_DEPENDS message_runtime gazebo_dev roscpp wave_gazebo_plugins
 )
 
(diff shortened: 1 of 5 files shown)


caguero:
I'm closing this contribution because VRX now relies on the upstream [Gazebo Thruster plugin](https://github.com/gazebosim/gz-sim/tree/gz-sim10/src/systems/thruster). If this functionality is still relevant, it should be implemented in the Gazebo Thruster plugin directly.

PR #482, opened 2022-07-19
Add CI for gazebosim

M1chaelM:
Very basic Galactic/fortress CI.

Looking around the internet, it seems that there are a few efforts underway to provide github actions to build and test gazebosim/ros2 projects. 
* For example, the Gazebosim repository uses this workflow: https://github.com/gazebosim/gz-sim/blob/ign-gazebo6/.github/workflows/ci.yml
* However, it doesn't seem to be functional right now, and the vast majority of PRs are being merged with failed checks.
* So, for now I think we should take a simple approach and just see whether the build succeeds with `colcon`.
* Later we can also add static checking.
* When the fancier build and test tools are working better we can switch over.

To test:
* Review the output details and verify that they say what we expect them to say.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,18 +10,27 @@ jobs:
       image: npslearninglab/watery_robots:galactic_current
     steps:
       - run: sudo chown -R `whoami`:`whoami` .
-      - uses: actions/checkout@v2.4.0
+      - name: Checkout vrx
+        uses: actions/checkout@v2.4.0
         with:
             ref: gazebosim
             path: vrx
 
-      - uses: actions/checkout@v2.4.0
+      - name: Checkout Gazebosim 
+        uses: actions/checkout@v2.4.0
         with:
             repository: ignitionrobotics/ros_ign
             ref: galactic 
             path: ros_ign
        
-     - name: Compile and test
-        id: ci
-        uses: ignition-tooling/action-ignition-ci@focal
+      - name: Set up workspace
+        run: |
+          mkdir -p /tmp/vrx_ws/src;
+          ln -s `pwd` /tmp/vrx_ws/src;
 
+      - name: Build and run tests
+        shell: bash
+        run: |
+          cd /tmp/vrx_ws;
+          source /opt/ros/galactic/setup.bash;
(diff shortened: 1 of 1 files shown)


M1chaelM:
@caguero I replaced the old line that said
```
catkin_make run_tests -j1
```
with
```
IGNITION_VERSION=fortress colcon test --merge-install --executor sequential;
```
Is that the right colcon command for testing our code? I notice that the output says two of our packages fail with `Ctest` errors, but we still get a green light, so not sure whether this is really working.

M1chaelM:
@caguero Do you have time to look at this today?

caguero:
> @caguero Do you have time to look at this today?

Yes, I'll review it today!

caguero:
> @caguero I replaced the old line that said
> 
> ```
> catkin_make run_tests -j1
> ```
> 
> with
> 
> ```
> IGNITION_VERSION=fortress colcon test --merge-install --executor sequential;
> ```
> 
> Is that the right colcon command for testing our code? I notice that the output says two of our packages fail with `Ctest` errors, but we still get a green light, so not sure whether this is really working.

That line looks correct, although I'd suggest to add `--event-handlers console_direct+` to get test output.

caguero (approved):
I wouldn't worry much about the tests for now because they're actually not for this project. I think we'd need to remove them, and check that the code compiles. Then, in separate PRs start adding tests and static checks.

M1chaelM:
@caguero OK, thanks! I was just writing something similar--it looks like the tests are failing for `ros_ign_bridge`, which is part of the dependency `ros_ign`, and not our project. So, I shouldn't be testing that. Agree?

caguero:
> @caguero OK, thanks! I was just writing something similar--it looks like the tests are failing for `ros_ign_bridge`, which is part of the dependency `ros_ign`, and not our project. So, I shouldn't be testing that. Agree?

Agree. Now that the basic CI is in, I'd focus on the tests that we can reuse for VRX.

caguero:
And when we have some tests, we should add the option to the colcon command for running only the tests inside the vrx package. 

```
--packages-select vrx_ros vrx_ign
```

PR #483, opened 2022-07-19
remove addition of zero Z valued grid point

acxz:
X.Z() never gets populated with a nonzero value besides it's initial construction value of 0. Thus this addition is extraneous.

Co-authored-by: Jonathan <jonathan.r.eastridge.civ@us.navy.mil>


caguero:
Closing in favor of #830 retargeted to `jazzy`.

Issue #484, opened 2022-07-20
Offload usv_dynamics buoyancy calculation to BuoyancyPlugin


acxz:
Follow up of #460 

It would great for readable and also extensibility if the buoyancy computation in the usv_dynamics_plugin (which is hardcoded for a two hull half cyclinder buoyant object) is replaced with the corresponding computation in the BuoyancyPlugin

Issue #167, opened 2019-09-22
Refining usv_gazebo_dynamics_plugin model


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/vrx-gh-pages/#!/osrf/vrx/issues/167)) by Anonymous.**

----------------------------------------
Hi, 

I noticed in the calculation of the buoyancy force in usv_gazebo_dynamics that the force saturates when deltaZ = hullRadius. 

However, in the wamv model, the pontoons are simulated as cylinders instead of half cylinders. Hence, would the calculation of buoyancy force saturate when the entire cylinder is submerged, instead of when half the cylinder is submerged? 

Separately, could I check what the parameter hullRadius refers to specifically? (i.e. radius of each pontoon)? 

This is because I noticed also that the buoyancy force calculates the upthrust for a single pontoon instead of 2 pontoons, as shown in the wamv model

M1chaelM:
This seems like it might be more of a bug than an enhancement, if we are not correctly implementing our model.

PR #485, opened 2022-07-20
Buoyancy plugin

acxz:
Solves #484 and #167 

The buoyancy computation code is removed from usv_dynamics and the generic BuoyancyPlugin is used.

The buoyant force has been verified to be within 50 N. Let me know if you would like videos for verification.


bsb808:
@acxz This is a really interesting idea.  When we wrote this plugin the standard Gazebo buoyancy plugin wasn't able to deal with items at the surface - only fully submerged items.

I tested the modification with an exaggerated sea-state.  There seems to be issues with synchronization with the visual wave field and lack of induced vessel pitch and roll.  Those are critical capabilities for our use-case.  

I'm not very familiar with the latest buoyancy.  Perhaps the standard plugin can be used for this application.  Is there documentation on how the plugin uses the <wave_model> and <buoyancy> tags?

bsb808:
Peek 2022-07-25 17-38

acxz:
@bsb808 thanks for looking at this PR! 

> I tested the modification with an exaggerated sea-state.

Can you provide the files you changed, so that I can test the scenario on my local machine.

> There seems to be issues with synchronization with the visual wave field and lack of induced vessel pitch and roll.

I believe this is a problem of the wave field visuals vs the actual simulated wave field height. I think you should also see similar behavior with the existing buoyancy code, without the plugin. Have you compared the two in this exaggerated scenario?

> Perhaps the standard plugin can be used for this application.

Here is the current upstream BuoyancyPlugin code: https://github.com/osrf/gazebo/blob/2e42eb91e7638da54b36dbb8fdd4ae5e3a310d6d/plugins/BuoyancyPlugin.cc

As you can tell, the upstream BuoyancyPlugin is not at all as fancy as the one we have here in VRX.

>  Is there documentation on how the plugin uses the <wave_model> and "buoyancy" tags?

For our plugin, the documentation is available here: https://github.com/osrf/vrx/blob/e5544ed275055d10ebffd78eb30fea9def962233/usv_gazebo_plugins/include/usv_gazebo_plugins/buoyancy_gazebo_plugin.hh#L79-L113

bsb808:
See https://github.com/osrf/vrx/wiki/changing_plugin_params_tutorial to change wave parameters.  I probably used period=10 and gain=1

acxz:
Thanks for sending the params you tested on. Just tested this PR and the existing codebase and the same behavior is observed.

For this PR in particular is there any closure that I can expect, now that the master branch has been put in maintenance mode.

cc'ing @M1chaelM since he has made issue #487 which this PR helps tackle.

caguero:
Following the spirit of the latest changes, we should use the upstream [Gazebo buoyancy system](https://github.com/gazebosim/gz-sim/tree/gz-sim10/src/systems/buoyancy) when possible. I suggest to use that plugin if you're still interested in this patch.

PR #489, opened 2022-07-23
Add mesh geometry to BuoyancyPlugin

### LINKED ISSUE #488: Add support for arbritary mesh geometries for buoyancy calculation
Currently, buoyancy is computed as an approximation of the object into a box, cylinder, or sphere. However, for actual models such an approximation can be limiting. It would be nice, if an arbitrary mesh file (which is already used for the visuals) can be used for the buoyancy computation.

acxz:
This allows the support for using a mesh file as the geometry used in the buoyancy computations. The buoyancy force computed from this PR is verified within 100N of the original buoyancy code. Which is close enough, to believe that this PR provides a more accurate buoyancy computation than the existing code.

Closes #488

### DIFF
--- a/usv_gazebo_plugins/CMakeLists.txt
+++ b/usv_gazebo_plugins/CMakeLists.txt
@@ -19,8 +19,8 @@ catkin_package(
   CATKIN_DEPENDS message_runtime gazebo_dev roscpp wave_gazebo_plugins
 )
 
-# Plugins require c++11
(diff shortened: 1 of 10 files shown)


caguero:
Following the spirit of the latest VRX changes, we should use the upstream [Gazebo buoyancy system](https://github.com/gazebosim/gz-sim/tree/gz-sim10/src/systems/buoyancy) when possible. I suggest to target this patch towards that plugin if you're still interested on it.

PR #491, opened 2022-07-26
Utilize proper pms coefficients

acxz:
The Pierson-Moskowitz uses a coefficient called $\beta$, which is defined as 0.74.
Source: https://wikiwaves.org/Ocean-Wave_Spectra

In our code, we have $\beta$ hardcoded as 5.0/4.0 = 1.25. This is incorrect.

This PR updates the Wavefield code to use the correct parameter.

Full equation: 

$$ S(\omega) = \dfrac{\alpha g^2}{\omega^5} \exp \left( - \beta * \left(\dfrac{\omega_0}{\omega}\right)^4 \right)$$

with $\alpha = 0.0081$ (correct in our code) and $\beta = 0.74$ (not correct in our code).

cc: @bsb808 @jeastnswc

### DIFF
--- a/usv_gazebo_plugins/CMakeLists.txt
+++ b/usv_gazebo_plugins/CMakeLists.txt
@@ -19,8 +19,8 @@ catkin_package(
   CATKIN_DEPENDS message_runtime gazebo_dev roscpp wave_gazebo_plugins
 )
 
-# Plugins require c++11
-set(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")
+# Plugins require c++17
+set(CMAKE_CXX_FLAGS "-std=c++17 ${CMAKE_CXX_FLAGS}")
 
 include_directories( include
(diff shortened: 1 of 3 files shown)


M1chaelM:
I'll let @bsb808 comment further, but it may be useful to refer to [our paper](https://wiki.nps.edu/display/BB/Publications?preview=/1173263776/1173263778/PID6131719.pdf) for the intended equations (see particularly equation 5 on page 3), and references.

acxz:
Sorry to not mention that. I did actually look at the papers and it the paper also uses the incorrect parameter, without citing the original paper.

acxz:
With the help of @jeastnswc, we have tracked down our confusion.

The original paper (Pierson and Moskowitz, 1964) lists the coefficient in spectral form which is a function of wind speed, while Frechot writes the equation in terms of the frequency of the spectral peak (which when simplified does not have the wind speed in the equation). This was clarified in (S.K. Chakrabarti, 1987) pg. 105

PR #496, opened 2022-08-03
Fluid density

### LINKED ISSUE #495: Inconsistant fluid density
Throughout the code, we have 3 or 4 different values that we use for the density of water. It would be best to consolidate the value of water density throughout the codebase.

I think it makes sense to use 1000 as the value to keep simulations relatively accurate but easy enough to work with.

acxz:
make consistent use of the density of water to 1000 kg/m^3.

closes #495


caguero:
Closing it in favor of #834 .

PR #486, opened 2022-07-21
Rmvanarse/gazebosim/scoring plugins dev

rmvanarse:
## Scoring plugin implemented with ignition

**Files added**
- `ScoringPlugin.hh`
- `ScoringPlugin.cc`
- `DummyPlugin.hh`
- `DummyPugin.cc`

The dummy plugins contain skeleton code that can be used for implementing other plugins. The scoring plugins have been structured similarly to the corresponding files in VRX. The main modifications are:
- Functions such as Load() and Update() have been replaced by Configure() and ISystemPreUpdate/ ISystemPostUpdate
- World/ model objects are now entities. Their attributes in Gazebo classic implementations are now components, accessed via the entity component system or a World/Model wrapper to the entities
- ROS publishers and subscribers have been replaced with ignition::transport and ROS messages have been replaced with ignition::msgs.
- Several functions for accessing or setting members have changed (e.g. GetLatitudeReference(), GetHeadingOffset() ). Some functions now require extra arguments such as the ECM or simTime (e.g. UpdateTime() ).
- Changes in data types - The main one is the replacement of ignition::common::Time (deprecated) with std::chrono::duration<double>. 

The following still needs to be done:
- Making time variables consistent in type or unit. Currently std::chrono::duration, std::chrono::steady_clock, ignition::msgs::Time and double are all being used. Additionally, ignition::common::Time, builtin_interfaces::msgs::Time and rclcpp::Time have been removed. 
- Completing the implementation of the PostUpdate function (Accessing the vehicle model and name)
- Code cleanup

PR #498, opened 2022-08-18
Working minimal ScoringPlugin example

j-herman:
This PR does not need to be merged - if you prefer, we can leave the branch as-is for reference.

This code implements a very basic version of the ScoringPlugin.  It loads the plugin, keeps track of the task time, and publishes changes in state (initial/ready/running/finished) to the /vrx/task/info topic.  Timing and some of messaging is stolen from Rishikesh's in-progress branch.  The goal is to provide some capability for testing new code segments in plugins that will derive from the base ScoringPlugin while full functionality is pending.

To test: start the simulation via `ros2 launch vrx_ros competition_local.launch.py ign_args:="-v 4 -r sydney_regatta.sdf"`
You should be able to see the ScoringPlugin listed under System Plugin info in the Component Inspector.
In a separate terminal window, echo the task topic via `ign topic --echo -t /vrx/task/info`
You should see the state message updating at 1Hz.  It will cycle through from initial to ready to running.  Finally, it will print "finished" once, and the simulation will exit.

Next steps: Ran into significant cmake problems trying to incorporate protobuf custom messages, and so the task message in this example is minimal.  Next major effort will be to resolve that issue.

### DIFF
--- a/vrx_ign/CMakeLists.txt
+++ b/vrx_ign/CMakeLists.txt
@@ -47,12 +47,26 @@ install(
   TARGETS Waves
   DESTINATION lib)
 
+# Base scoring plugin
+add_library(ScoringPlugin SHARED
+  src/ScoringPlugin.cc
+)
+target_link_libraries(ScoringPlugin PUBLIC
+ignition-gazebo${IGN_GAZEBO_VER}::core
+ignition-plugin${IGN_PLUGIN_VER}::ignition-plugin${IGN_PLUGIN_VER}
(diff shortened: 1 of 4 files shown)


M1chaelM:
This runs fine for me and I see the behavior described. We should discuss whether to use this or PR #486 as a basis for moving forward.

j-herman:
Made some modifications to better support inheritance based on work on the Stationkeeping scoring plugin - need to build this library first.  Changed a few function declarations that shouldn't be required overrides, and fixed initialization of timing variables.

caguero:
Replaced with #501.

PR #502, opened 2022-08-26
Jessica/stationkeeping2

j-herman:
To test: start the simulation via `ros2 launch vrx_ros competition_local.launch.py ign_args:="-v 4 -r sydney_regatta.sdf"` 
You should be able to see the StationkeepingScoringPlugin listed under System Plugin info in the Component Inspector.
In a separate terminal, launch the wamv: `ros2 launch vrx_ign spawn.launch.py name:=wamv world:=sydney_regatta model:=wam-v x:=-532 y:=162 z:=0.1 R:=0 P:=0 Y:=1
`
In a third terminal, echo the goal topic: `ign topic --echo -t /vrx/stationkeeping/goal`
You should see the goal message updating at 1Hz during the "ready" and "running" states.
Next, echo the pose error and/ or mean pose error topics: 
`ign topic --echo -t /vrx/stationkeeping/pose_error`
`ign topic --echo -t /vrx/stationkeeping/mean_pose_error`
You should see the error message updating at 1Hz in the "running" state.  
Move the WAMV and note that the value of the error messages changes.
Echo the task info topic from the main ScoringPlugin to see the score updating: `ign topic --echo -t /vrx/task/info`

Lat/long and gz coordinates have been verified and scoring looks like it is working correctly.  There's a small position conversion error that results in a score of ~0.5 (ideally zero) when the goal pose is entered as a lat/long.  I think this is as expected and similar to what we had in the old simulation.
Per the discussion, I've disabled the Markers code for now and will either replace it with the old waypoint_markers class or get it working in a future PR.


M1chaelM:
@caguero @j-herman For some reason I'm having a lot of trouble running ignition on this simulation. It starts out ok but then grinds to a halt after a few seconds. Ubuntu reports that the application is unresponsive and asks whether I'd like to wait or force quit.

I suspect this is a local problem, but if anyone has suggestions for troubleshooting I'd appreciate it.

M1chaelM:
> Shoot. That started happening for me as well - something changed upstream, I think, in the defaults for MarkerManager. I commented out the line to use the markers in the sdf and all seems fine now (you need to actually remove the line, not just set it to false). @M1chaelM

The last commit resolved the problem for me. I can now run this demo and see the `/vrx/task/info` topic, which seems to be working correctly. It shows the task name and the various time values, and score (which looks reasonable).

I'm not getting any output for the 3 task-specific topics:
* `/vrx/stationkeeping/goal`
* `/vrx/stationkeeping/pose_error`
* `vrx/stationkeeping/mean_pose_error`

At the moment I don't see any messages in my output explaining why this might be the case. Can you verify that you are able to see messages on these topics? I'll try to look at the source and see if anything jumps out.

j-herman:
My mistake.... Missing underscore in the topics.  I've just updated the comment - should be "/vrx/station_keeping/goal", etc.

caguero (approved):
I made a few style tweaks and verified that it's working!

PR #503, opened 2022-08-30
WAM-V joystick teleoperation

caguero:
Requires:
  * https://github.com/gazebosim/gz-sim/pull/1681

This pull request adds the ability to teleoperate the WAM-V, which I think it's very useful for testing scoring plugins.

As a reminder, you need a Logitech F310 controller with the back switch in `D` mode.

### How to test it?

Launch the simulation:
```
ros2 launch vrx_ros competition_local.launch.py ign_args:="-v 4 -r sydney_regatta.sdf"
```

Launch the WAM-V:
```
ros2 launch vrx_ign spawn.launch.py name:=wamv world:=sydney_regatta model:=wam-v x:=-532 y:=162 z:=0 R:=0 P:=0 Y:=1
```

Launch the `usv_joy_teleop` launch file:
```
ros2 launch vrx_ign usv_joy_teleop.py
```

To control the WAM-V you have to keep the `L1` button pressed (deadman button). Other than that, the controls are the same that we used to have. The ball shooter is not installed in the WAM-V but you can confirm that the message is sent by echoing the ` /wamv/shooters/ball_shooter/fire` topic while you pressed the `A` button.
```
ros2 topic echo /wamv/shooters/ball_shooter/fire
data: true
---
```

**Note**: I think that the [joy_teleop package](https://github.com/ros-teleop/teleop_tools/tree/foxy-devel/joy_teleop) is still not polished enough to be used with ROS2. As far as I know it's not possible to configure it without a deadman button, which I think it's a bug.


M1chaelM:
@caguero it looks like testing this PR currently requires building gazebosim from source. Is that right? I see you've merged #1681 but I'm not sure how long this change takes to propagate to the binaries.

M1chaelM (approved):
This seems to be working as I would expect. A minor caveat is that the thrusters seem more powerful than I remember. For example, by switching from full-forward to full-back, I can stop almost instantly. I can also spin fast in a tight circle, then reverse to spin in the opposite direction with very little transition time. I don't think this is necessarily a problem but it does seem different. Maybe it's a result of not having any of the components on the wamv yet?

caguero:
> This seems to be working as I would expect. A minor caveat is that the thrusters seem more powerful than I remember. For example, by switching from full-forward to full-back, I can stop almost instantly. I can also spin fast in a tight circle, then reverse to spin in the opposite direction with very little transition time. I don't think this is necessarily a problem but it does seem different. Maybe it's a result of not having any of the components on the wamv yet?

This is intended because I set `500` as the scale in the `wamv.yaml`. I thought it would be handy for testing but let me know if that was too much.

PR #513, opened 2022-09-17
Navigation scoring plugin port

M1chaelM:
Port of the navigation scoring plugin to new Gazebosim / ros2. 

## Requirements
* Gazebosim must be installed from source.
* Gamepad is needed for WAMV teleop.

## Preparation
The following adjustments must be made to the individual repositories within the gazebo workspace before building:

### gz-fuel-tools
* Make sure `gz-fuel-tools` is checked out to commit `7b6ce8a4ab4cbfd724eefba7f2e57892671abfc0`
  ```
  cd gazebo_ws/src/gz-fuel-tools
  git checkout 7b6ce8a4ab4cbfd724eefba7f2e57892671abfc0
  ```
* Previous commits may work, but the 2 subsequent commits have a bug that will cause a runtime crash.

### enable teleoperation
* We need to teleoperate the WAMV to make sure the gates are doing the right thing. This requires functionality that has not yet been merged from v6 to v7, so we need to use the `6-to-7` branch:
  ```
  cd gazebo_ws/src/gz-sim
  git checkout 6-to-7-20221021
  ```
### rebuild
After making the above changes, rebuild gazebo:
```
cd ~/gazebo_ws
colcon build --merge-install
```
Some warnings are expected. Building a second time should produce no stderr output.

## Build
In a separate terminal, checkout the `M1chaelM/navigation_scoring_plugin` branch and build it:
```
cd ~/vrx_ws/src/vrx
git checkout M1chaelM/navigation_scoring_plugin
cd ~/vrx_ws
```
Again, some warnings are expected for the first build.

## Run
Open three new terminals. In each, source the Gazebo and VRX workspaces:
```
cd vrx_ws
source ~/gazebo_ws/install/setup.bash
. install/setup.bash
```
Run the simulation in one terminal:
```
ros2 launch vrx_gz competition.launch.py world:=navigation_task
```
Run teleop in a second terminal:
```
```
Echo the `task/info` topic in the third terminal:
```
```

## Test
* Wait for the task to enter `onrunning` state.
* Use the gamepad to drive the WAMV through the gates in the course.
* Check that a "New gate crossed!" message is printed to terminal each time a gate is crossed.
* Check that the simulation exits after all 3 gates are crossed.
* Check that the score printed in the `vrx/task/info` message is reasonable. It should be equal to the elapsed time between the crossing the entrance gate the exit gate.
* Check that crossing a gate out of order terminates the task.
* Check that crossing a gate in the wrong direction terminates the task.

## Known issues
* The WAMV does not lock, so it is possible to drive through the first gate before the simulation enters the `onrunning` state. This will cause the gate to be ignored. Passing a second gate will then be treated as crossing a gate out of order, which terminates the simulation.
* Currently the entire simulation terminates the moment the task is over. In Gazebo Classic, my memory is that it stayed up but  just froze in place.
* If we use this task again we may want to revisit how we calculate the score. Starting the timer at the entrance gate seems odd.


M1chaelM:
Looks like the relative position of the gate is not being calculated correctly in the `IsPoseInGate` method, here:
```
const math::Vector3d robotLocalPosition =
    this->pose.Rot().Inverse().RotateVector(_robotWorldPose.Pos() -
    this->pose.Pos());
```
Either this transformation is not working for some reason, or the gate pose is wrong (or the vehicle pose, but this seems less likely).

M1chaelM:
It appears that the problem is related to our nested course model. The function:
```
auto leftMarkerPose = _ecm.Component<gazebo::components::Pose>(
    leftMarkerEntity)->Data();
```
returns the relative pose of the marker inside the course, rather than the pose in the world frame. I assume there is a way to ask for the pose relative to the world instead. It looks like this should be simply:
```
auto leftMarkerPose = _ecm.Component<gazebo::components::WorldPose>(
    leftMarkerEntity)->Data();
```
This compiles, but when I go to run it the simulation mysteriously shuts down right away.

@caguero or @j-herman do you know what's going on here? Or what is the right way to get the world pose of a nested model?

j-herman:
@M1chaelM I am accessing world pose info for nested models via the Link wrapper rather than directly via the ECM.  I don't have that part of the code working yet, so I can't verify that it gives good info, but it doesn't crash.  That might be an option, although your code looks like it should work...  gazebo::components::WorldPose is defined right along with gazebo::components::Pose, and so I would think that you should be able to access the component with the same code. (https://github.com/gazebosim/gz-sim/blob/ign-gazebo6/include/ignition/gazebo/components/Pose.hh)
You could also try an alternate function.  This comes from Link.cc:
`_ecm.ComponentData<components::WorldPose>(this->dataPtr->id);`
It returns std::optional<math::Pose3d> so you'd need to extract the information.

M1chaelM:
This is now working and is almost ready to be ported over to Garden. I have 3 hopefully quick questions to resolve first:
* The ["New Gate Crossed"](https://github.com/osrf/vrx/pull/513/files#diff-643fb64f8f1f8fa95b7f8b3922bf369fd62b1bb61fa480c3d869839d46b330a1R414) message, which is printed using `ignmsg << "New Gate Crossed` does not display until the plugin has exited. Is there some flush command or something needed to get it to print immediately?
* I'm confused about the the [`silent` attribute of the NavigationScoringPlugin class](https://github.com/osrf/vrx/pull/513/files#diff-643fb64f8f1f8fa95b7f8b3922bf369fd62b1bb61fa480c3d869839d46b330a1R130). Do we need this or should it be part of the parent ScoringPlugin class instead?
* I'm not sure what [the `GZ_ASSERT` macro was doing here](https://github.com/osrf/vrx/pull/513/files#diff-643fb64f8f1f8fa95b7f8b3922bf369fd62b1bb61fa480c3d869839d46b330a1R237). Is there a version that works for the new Gazebo, and do we need it?

@caguero If you have a moment could you take a look at the above questions? I think they are all relatively minor but just want to get them resolved.

caguero:
> This is now working and is almost ready to be ported over to Garden. I have 3 hopefully quick questions to resolve first:
> 
> * The ["New Gate Crossed"](https://github.com/osrf/vrx/pull/513/files#diff-643fb64f8f1f8fa95b7f8b3922bf369fd62b1bb61fa480c3d869839d46b330a1R414) message, which is printed using `ignmsg << "New Gate Crossed` does not display until the plugin has exited. Is there some flush command or something needed to get it to print immediately?
> * I'm confused about the the [`silent` attribute of the NavigationScoringPlugin class](https://github.com/osrf/vrx/pull/513/files#diff-643fb64f8f1f8fa95b7f8b3922bf369fd62b1bb61fa480c3d869839d46b330a1R130). Do we need this or should it be part of the parent ScoringPlugin class instead?
> * I'm not sure what [the `GZ_ASSERT` macro was doing here](https://github.com/osrf/vrx/pull/513/files#diff-643fb64f8f1f8fa95b7f8b3922bf369fd62b1bb61fa480c3d869839d46b330a1R237). Is there a version that works for the new Gazebo, and do we need it?
> 
> @caguero If you have a moment could you take a look at the above questions? I think they are all relatively minor but just want to get them resolved.

`GZ_ASSERT` is defined in `gz-common` with `#include <gz/common/Util.hh`. It works the same way as in Gazebo classic. If the condition isn't true, execution terminates with the associated message. Note that `asserts` are ignored when the code is compiled in release mode.

I see the `silent` attribute in the base scoring plugin already. It looks like we could remove all these print messages and the `silent` member from the `Navigation` plugin.

I'm actually not sure about why `gzmsg` doesn't flush the messages...

iche033:
> The "New Gate Crossed" message, which is printed using ignmsg << "New Gate Crossed does not display until the plugin has exited. Is there some flush command or something needed to get it to print immediately?

I think this may actually make it flush and output to the console:

```cpp
ignmsg << "some msg" << std::endl;
std::cout << std::flush;
```

caguero:
> @caguero and @iche033 Somewhere in the switch to Garden I lost the ability to drive the WAMV. I've tried the following:
> 
> ```
> ros2 launch vrx_gz usv_joy_teleop.py
> ```
> 
> This runs and seems to be responding to the gamepad, but in an unexpected way. All I can do is change the angle of the thrusters. I can't apply thrust.
> 
> ```
> ros2 topic pub /wamv/thrusters/left/thrust std_msgs/msg/Float64 "data: 1.0"
> ```
> 
> This runs without error but the WAMV doesn't respond.
> 
> ```
> gz topic --pub "data: 1.0" --topic /wamv/thrusters/left/thrust --msgtype Double
> ```
> 
> This also runs without error but the WAMV doesn't respond.
> 
> Do you have any suggestions for troubleshooting?

That's expected until we forward port https://github.com/gazebosim/gz-sim/pull/1681 to Garden.

M1chaelM:
> > @caguero and @iche033 Somewhere in the switch to Garden I lost the ability to drive the WAMV. I've tried the following:
> > ```
> > ros2 launch vrx_gz usv_joy_teleop.py
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This runs and seems to be responding to the gamepad, but in an unexpected way. All I can do is change the angle of the thrusters. I can't apply thrust.
> > ```
> > ros2 topic pub /wamv/thrusters/left/thrust std_msgs/msg/Float64 "data: 1.0"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This runs without error but the WAMV doesn't respond.
> > ```
> > gz topic --pub "data: 1.0" --topic /wamv/thrusters/left/thrust --msgtype Double
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This also runs without error but the WAMV doesn't respond.
> > Do you have any suggestions for troubleshooting?
> 
> That's expected until we forward port [gazebosim/gz-sim#1681](https://github.com/gazebosim/gz-sim/pull/1681) to Garden.

Ah, OK. I think we should probably do that before merging this PR then. Otherwise it is difficult to test that we are correctly detecting when the WAMV passes through a gate.

caguero (approved):
It works for me as expected!

M1chaelM:
> I took the liberty of making some minor changes in [b10ae43](https://github.com/osrf/vrx/pull/513/commits/b10ae430cc1a325a0e8343db780ca81e0ce9b77d).

Thanks! If we have time I'd like to ask you about a few of these tomorrow.

Issue #514, opened 2022-09-22
Clean up the infrastructure from MBZIRC


caguero:
To bootstrap the migration to Gazebo we took some of the ROS 2 infrastructure from the [MBZIRC project](https://github.com/osrf/mbzirc). However, there are still some leftovers from MBZIRC that we don't need or they don't apply in VRX. We need to go over the project files and refactor a bit that infrastructure code.

caguero:
Done in multiple iterations.

PR #519, opened 2022-09-28
[Garden] Remove code for spawning arm, gripper, and fixed wing

iche033:
* Removed code in bridge and spawn scripts related to arm, gripper, and fixed wing
* Removed usv_bridge: this is MBZIRC-specific. It was used to provide hardware interface topics for controlling USV
* Removed stream_status topic: this is also MBZIRC-specific
* Removed `mavros_msgs` dependency in cmake as it's not currently used

I'm able to control the wam-v through gz topics, e.g.

```sh
gz topic -t "/wamv/thrusters/right/pos" -m ignition.msgs.Double -p "data: -1.0"
gz topic -t "/model/wamv/joint/left_engine_propeller_joint/cmd_thrust" -m ignition.msgs.Double -p "data: 50.0"
```

However, the ROS2 topics are not working for me (in `gazebosim` branch as well). Looks like the topics being bridged are not correct. I'll fix this in a follow up PR


caguero:
Part of #514 .

PR #520, opened 2022-09-28
[Garden] Fix bridging Gz to ROS2 topics

iche033:
Signed-off-by: Ian Chen <ichen@openrobotics.org>

Follow up to #519, this PR fixes the ROS2 topics for the wamv. The problem was that the [ros_ign_bridge](https://github.com/gazebosim/ros_gz/tree/humble/ros_gz_bridge) is still using the `ignition.msgs.*` msg types instead of `gz.msgs` (see msg type table `README.md`) , presumably to keep compability with Fortress. 

All ROS2 topics should be working except the thruster cmd topic. It requires this https://github.com/gazebosim/gz-sim/pull/1681 to be forward ported from Fortress to Garden

The thrusters joint pos topic should work. 
```
ros2 topic pub': ros2 topic pub --once wamv/thrusters/right/pos std_msgs/msg/Float64 'data: -1
```
I removed more topic bridges that are not needed / used:
* `/wamv/imu/data` - commented out as there is no imu sensor on the wamv yet
* `/wamv/rx` and `/wamv/tx` - commented out as RFComms is not being used yet
* `/vrx/score`, `/vrx/run_clock`, `/vrx/phase`: removed as these are also MBZIRC-specific. From looking at the vrx [technical guide](file:///home/ian/Desktop/VRX2022_Technical-Guide_v1.1.pdf), I think there is a `/vrx/task/info` topic with custom message type that captures these status data instead.

### DIFF
--- a/vrx_gz/src/vrx_gz/bridges.py
+++ b/vrx_gz/src/vrx_gz/bridges.py
@@ -10,7 +10,7 @@ def imu(world_name, model_name, link_name='base_link'):
     return Bridge(
         gz_topic=f'{sensor_prefix}/imu_sensor/imu',
         ros_topic='imu/data',
-        gz_type='gz.msgs.IMU',
+        gz_type='ignition.msgs.IMU',
         ros_type='sensor_msgs/msg/Imu',
         direction=BridgeDirection.GZ_TO_ROS)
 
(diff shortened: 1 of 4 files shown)


caguero (approved):
Works for me! Should we also bridge the vrx task message. It's a `msgs::Param`. We should be able to use the `ros_gz_interfaces/ParamVec`.

iche033:
> Works for me! Should we also bridge the vrx task message. It's a `msgs::Param`. We should be able to use the `ros_gz_interfaces/ParamVec`.

yeah, I can work on that in a separate PR

PR #522, opened 2022-09-29
[Garden] Bridge task topics

iche033:
Bridged a few more topics from gz to ROS2:

```
/vrx/contacts
/vrx/stationkeeping/goal
/vrx/stationkeeping/mean_pose_error
/vrx/stationkeeping/pose_error
/vrx/task/info
```

Data should be printed when you do a `ros2 topic echo` on these topics. For `/vrx/contacts`, I tested by manually moving the wamv to collide with the buoys.

### DIFF
--- a/vrx_gz/src/vrx_gz/bridges.py
+++ b/vrx_gz/src/vrx_gz/bridges.py
@@ -109,3 +109,43 @@ def clock():
         gz_type='ignition.msgs.Clock',
         ros_type='rosgraph_msgs/msg/Clock',
         direction=BridgeDirection.GZ_TO_ROS)
+
+def task_info():
+    return Bridge(
+        gz_topic=f'/vrx/task/info',
+        ros_topic=f'/vrx/task/info',
+        gz_type='ignition.msgs.Param',
+        ros_type='ros_gz_interfaces/msg/ParamVec',
+        direction=BridgeDirection.GZ_TO_ROS)
+
+def contacts():
+    return Bridge(
+        gz_topic=f'/vrx/contacts',
+        ros_topic=f'/vrx/contacts',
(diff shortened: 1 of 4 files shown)


caguero (approved):
Works for me! I fixed a small previous bug in [80260e9](https://github.com/osrf/vrx/pull/522/commits/80260e9ec37a189bcf2d871c0e3f6f62502a1725).

The station keeping topics are only relevant when the station keeping task is running. Is there any way to enable these topics only when the `StationKeepingScoringPlugin` is running?

iche033:
> The station keeping topics are only relevant when the station keeping task is running. Is there any way to enable these topics only when the StationKeepingScoringPlugin is running?

When does the station keeping task start running? Do we want to hide topics completely and launch the bridge process only when the task starts?

caguero:
> > The station keeping topics are only relevant when the station keeping task is running. Is there any way to enable these topics only when the StationKeepingScoringPlugin is running?
> 
> When does the station keeping task start running? Do we want to hide topics completely and launch the bridge process only when the task starts?

Oh, I meant that there might be some worlds where the station keeping plugin isn't running. Then, we shouldn't see any station keeping topics.

iche033:
ohh I see. One idea is to keep track of the worlds in the python script and only bring up the bridges if that world has station keeping task. I can take a look at this

caguero:
> ohh I see. One idea is to keep track of the worlds in the python script and only bring up the bridges if that world has station keeping task. I can take a look at this

I think that will work, we'll need to modify the worlds list when running the competitions because we use extra worlds that aren't public, but I think it sounds reasonable.

caguero:
Part of #514 .

PR #530, opened 2022-10-13
[Garden] Fix crash on when resuming simulation from paused state

iche033:
Signed-off-by: Ian Chen <ichen@openrobotics.org>

Gazebo currently crashes if you pause then resume simulation. This is due to the `PolyhedraBuoyancyDrag` setting invalid link forces when `dt` is zero. The changes here just tell the plugin to skip computation when simulation is paused

To reproduce the crash:

1. Launch sim with any world (all of them should have buoys with the `PolyhedraBuoyancyDrag` plugin). 

1. Pause simulation

1. Play simulation -> crash


caguero (approved):
Works for me!

PR #537, opened 2022-11-09
[Garden] Port scan, dock, deliver - Part 1

caguero:
This pull request partially ports the `ScanDockScoringPlugin` from Gazebo classic. In particular, it ports the functionality related with the color sequence checker.

### How to test it?

* Launch the scan_dock_deliver_task world:
```
ros2 launch vrx_gz competition.launch.py world:=scan_dock_deliver_task
```

* On a new terminal check the `/vrx/task/info` topic:
```
ros2 topic  echo /vrx/task/info
```

* Wait until we enter into `running` state (around 20 secs), and then, submit a correct color sequence. On a new terminal launch:
```
ros2 topic pub /vrx/scan_dock_deliver/color_sequence --once ros_gz_interfaces/msg/StringVec "header:
  stamp:
    sec: 0
    nanosec: 0
  frame_id: ''
data: ["red", "green", "blue"]"
```
Verify in your previous terminal that you get 10 points.

* Additionally, you can rerun the simulation and submit an incorrect sequence. You shouldn't get any points. Additional submissions should be ignored, as there's only one attempt.


vrx_gz/src/ScanDockScoringPlugin.cc
```diff
@@ -0,0 +1,311 @@
...
+void ColorSequenceChecker::OnColorSequence(const msgs::StringMsg_V &_sequence)
+{
+  gzmsg << "ColorSequenceChecker: Color sequence submission received:"
+        << _sequence.DebugString() << std::endl;
+  std::cout << std::flush;
+
+  // Sanity check: Only one submission is allowed.
+  if (this->colorSequenceReceived)
+  {
+    gzerr << "The color sequence has already been submitted" << std::endl;
```
M1chaelM, line 124: This does not seem to be displaying, despite the fact that the variable gets set to true in line 137, below. Maybe the problem is that the color checker gets disabled and stops listening for messages.

vrx_gz/src/ScanDockScoringPlugin.cc
```diff
@@ -0,0 +1,311 @@
...
+    return;
+  }
+
+  // Sanity check: We should receive three colors.
+  if (_sequence.data_size() < 3u)
```
M1chaelM, line 129: Currently, if we submit a sequence with more than 3 colors, it will just score the first three. I'm wondering if we should change `< 3u` to `!= 3u` to also discard sequences that are too long.

M1chaelM (approved):
Two very small suggestions. Otherwise this is working as expected.

vrx_gz/src/ScanDockScoringPlugin.cc
caguero: It's a bit redundant because as you said, it's disabled when the first submission is processed.

vrx_gz/src/ScanDockScoringPlugin.cc
caguero: Sounds good to me. Changed in [558ef4d](https://github.com/osrf/vrx/pull/537/commits/558ef4d29f251d7756f965fd462c5404cc86305c).

PR #541, opened 2022-11-17
[Garden] Port scan, dock, deliver - Part 2

caguero:
This pull request partially ports the ScanDockScoringPlugin from Gazebo classic. In particular, it ports the functionality related with the dock checker.

### How to test it?

* Launch the scan_dock_deliver_task world:
```
ros2 launch vrx_gz competition.launch.py world:=scan_dock_deliver_task
```

When the simulation transitions to the `ready` state (~10secs), you should see how the placards update their symbols to the ones specified in SDF (yellow circle, blue triangle and red rectangle).

* On a new terminal launch the joystick teleop node:
```
```

* On a new terminal check the `/vrx/task/info` topic:
```
ros2 topic  echo /vrx/task/info
```

* Now, teleop the WAM-V to a bay. A successful docking maneuver consists on staying for around `<min_dock_time>` (10 seconds) in a docking bay, and then, leave. Verify that you get `15` points for docking in any bay and `5` extra points if you dock in the correct bay (the central one). Keep also in mind that the simulation will terminate after any successful docking.

**Bonus**: I reduced the sensitivity of the joystick teleoperation.


M1chaelM (approved):
I made some minor superficial changes to variable names to align with the fact that docking in the wrong bay is still "allowed" but just doesn't get extra points.

Everything is working correctly but because the simulation exits so quickly the score in `/vrx/task/info` does not have time to update. Maybe another reason to allow the simulation to stay up for a bit after the task is over?

caguero:
> os2 topic  echo /vrx/task/info

I just verified that the score is correct and it seems so:

```
[ruby $(which gz) sim-1] [Msg] [ScanDockScoringPlugin.cc:720] Score: 20
```

I'm not opposed to delay the termination, we can consider it in a separate PR.

PR #531, opened 2022-10-14
[Garden] Port AcousticPinger plugin

caguero:
This pull request ports the `AcousticPingerPlugin` to Gazebo.

### How to test it?

1. Run the simulation loading the new `gymkhana_task` world:
```
ros2 launch vrx_gz competition.launch.py world:=gymkhana_task
```

2. On a separate terminal echo the `pinger` topic:
```
ros2 topic echo /wamv/pingers/pinger/range_bearing
```

The pinger is located 2 meters below the red buoy. You should get a range of around 29m. You can also try rotating/translating the WAM-V to see how the angles change. Keep in mind that there's some noise applied.

3. You can also try changing the pinger position (for example to the origin):
```
ros2 topic pub /wamv/pingers/pinger/set_pinger_position --once geometry_msgs/msg/Vector3
```

4. Verify how the range changes accordingly.

PR #526, opened 2022-10-07
[Garden] Port WAMV-V xacro files and generation script

iche033:
* Created a new `vrx_urdf` directory that consists of these packages:
  * `wamv_description` - Ported from `main` and updated xacro files to to use new gazebo systems
  * `wamv_gazebo` - Ported from `main` and updated xacro files (incl. `components` and `thruster_layout`) to use new gazebo systems and sensors
  * `vrx_gazebo` - ported the wam-v generation python and launch scripts to ROS2
* Updated the launch script in `vrx_gz` to support loading a urdf file of the wam-v
  * i.e. added `urdf` arg to `competition.launch.py`
  * The launch scripts convert urdf -> sdf before spawning the wam-v into simulation
* Gz<->ROS2 bridges for these new components / sensors should be working
  * `python-sdformat` is used to parse the generated sdf string in order to determine what sensors are attached to the wam-v and then launch the corresponding bridges. Note if you built sdformat from source, you will need to set the `PYTHONPATH`. See test instructions below.

To test:

1. Run the `generate_wamv` script to generate a urdf file from the provided example configuration yaml files.

    ```sh
    ros2 launch vrx_gazebo generate_wamv.launch.py component_yaml:=`ros2 pkg prefix vrx_gazebo`/share/vrx_gazebo/config/wamv_config/example_component_config.yaml thruster_yaml:=`ros2 pkg prefix vrx_gazebo`/share/vrx_gazebo/config/wamv_config/example_thruster_config.yaml wamv_target:=`pwd`/wamv_target.urdf wamv_locked:=False
    ```
    This should output a `wamv_target.urdf` in your current working directory.

1. Run the simulator with the generated wamv file

    a. If you installed sdformat from source, you will need to export your `PYTHONPATH`
        
       export PYTHONPATH=$PYTHONPATH:<path_to_colcon_ws>/install/lib/python 

    b. Launch sim

       ros2 launch vrx_gz competition.launch.py world:=sydney_regatta urdf:=`pwd`/wamv_target.urdf

vrx_urdf

```sh
ros2 topic list
...

/clock
/parameter_events
/rosout
/tf
/tf_static
/vrx/contacts
/vrx/task/info
/wamv/pose
/wamv/pose_static
/wamv/sensors/cameras/far_left_camera_sensor/camera_info
/wamv/sensors/cameras/far_left_camera_sensor/image_raw
/wamv/sensors/cameras/far_left_camera_sensor/optical/camera_info
[...]

PR #543, opened 2022-11-22
[caguero] Add pinger plugin to wamv_pinger.xacro

caguero:
After #526 , the pinger plugin wasn't working. This pull request adds the plugin to the xacro macro.

For testing, follow the instructions detailed in #531 .


M1chaelM:
I'm seeing output now, but the range is around 42 rather than 29. Is that OK?

M1chaelM (approved):
As mentioned in the other comment, this gives me a range of around 42 rather than 29 (which is what is listed in the testing instructions). However, after moving the pinger to the origin, I was able to verify that I get a range of around 524, which makes sense for the wam-v's coordinates of about -486, 197.

PR #550, opened 2022-12-07
[Garden] Delay the simulation exit

caguero:
This pull request modifies the way we exit the simulation. Before, the simulation would terminate when:
  * `<per_plugin_exit_on_completion>` is set to `true` and the `remaining time` reaches `0`, or
  * `VRX_EXIT_ON_COMPLETION` environment variable is set to `true` and the `remaining time` reaches `0`, or
  * A plugin calls `Exit()`.
 
Now, the behavior is to terminate the simulation `<exit_time>` seconds after:
  * `<per_plugin_exit_on_completion>` is set to `true` and the `remaining time` reaches `0`, or
  * `VRX_EXIT_ON_COMPLETION` environment variable is set to `true` and the `remaining time` reaches `0`, or
  * A plugin calls `RequestExit()`.

~~I replaced `Exit()` with `RequestExit()` to make it a bit more explicit that you're requesting the termination but that won't happen right away.~~

Plugins should call `Finish()` to request the finish of the task. If needed, the simulation will also be terminated after the delay.

How to test it?

* Run the perception task:
```
ros2 launch vrx_gz competition.launch.py world:=perception_task
```

* Leave the simulation running and observe that simulation should terminate at second `42.0` (40 + 2).

### DIFF
--- a/vrx_gz/src/GymkhanaScoringPlugin.cc
+++ b/vrx_gz/src/GymkhanaScoringPlugin.cc
@@ -211,19 +211,6 @@ void GymkhanaScoringPlugin::PreUpdate(const gz::sim::UpdateInfo &_info,
 }
 
 //////////////////////////////////////////////////
-void GymkhanaScoringPlugin::OnRunning()
-{
-  if (!this->dataPtr->silent)
-  {
-    gzmsg << "GymkhanaScoringPlugin::OnRunning" << std::endl;
-    std::cout << std::flush;
-  }
-  ScoringPlugin::OnRunning();
-}
-
-
-
-//////////////////////////////////////////////////
 void GymkhanaScoringPlugin::OnFinished()
 {
   // TODO: fix obstacle penalty (Navigation Plugin Implementation)
--- a/vrx_gz/src/GymkhanaScoringPlugin.hh
+++ b/vrx_gz/src/GymkhanaScoringPlugin.hh
@@ -87,9 +87,6 @@ namespace vrx
     public: void ChannelCallback(const gz::msgs::Param &_msg);
 
     // Documentation inherited.
-    private: void OnRunning() override;
- 
-    // Documentation inherited.
     protected: void OnFinished() override;
  
     /// \brief Private data pointer.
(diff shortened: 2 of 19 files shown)


M1chaelM:
This works fine and solves the problem of the final score not being displayed/recorded by the `/vrx/task/info` message. However, I'm a little confused about our various ways of finishing the plugins, so I want to summarize to make sure I understand. 

We now have the following related functions:
* `ScoringPlugin::RequestExit`
  * Changes task state to "finished", sets remaining time to 0, sets exitTime to current time plus delay.
* `ScoringPlugin::Exit`
  * Checks whether plugin should close Gazebo when it finishes, and then either closes Gazebo or prints a message.
* `ScoringPlugin::Finish`
  *  Checks whether task state is "finished". If not, sets it to "finished" and calls `OnFinished`
* `ScoringPlugin::OnFinished`
  * Prints a message (if not silent), sets score to `timeoutScore` if specified, updates and publishes `taskMsg`.

How these work together:
* `Exit` is called from `UpdateTaskState` if `taskState` is "finished" and `currentTime` >= `exitTime`
* `UpdateTaskState` is called in `PreUpdate`
* `taskState` can be changed to "finished" under the following circumstances:
  * By `UpdateTaskState` if `taskState` is "running" and `timedOut` is `true`.
  * By `Finish`, whenever it's called.
  * By `RequestExit`, whenever it's called.
* `OnFinished` is called under the following circumstances
  * By `UpdateTaskState` if it transitions to the `finished` state.
  * By `Finish` whenever it is called.

caguero:
> 03bd37e

Thanks for the suggestion. I agree that the logic could be more clear. I simplified in [03bd37e](https://github.com/osrf/vrx/pull/550/commits/03bd37e9fa9df30ee99339d3272b36573a863d65).

M1chaelM:
> Thanks for the suggestion. I agree that the logic could be more clear. I simplified it in [03bd37e](https://github.com/osrf/vrx/pull/550/commits/03bd37e9fa9df30ee99339d3272b36573a863d65).

Thanks! These changes look good to me. To recap: 
* we are eliminating `RequestExit` and using `Finish` instead. 
* `Finish` has been modified to set the `exitTime` to the current time plus the exit delay. 
* We are no longer setting the remaining time to `0`. I think this is because there is nothing relying on that value, and it's fine for our task message to display what the remaining time was when the task entered the finished state.

M1chaelM (approved):
Works as described.

Issue #545, opened 2022-11-23
WAMV is not locking before OnRunning


M1chaelM:
For many of the tasks ported to gazebosim, the WAMV does not lock properly and can move during the OnReady state. This is definitely the case for the navigation and gymkhana tasks. I will check whether the lock is working for any tasks and update this issue.

caguero:
I've just tested the latest `gazebosim` branch and the WAM-V loks/unlocks as expected in all the task worlds. The only world that wasn't working as expected was `sydney_regatta`, as the USV never unlocked. #549 should fix this issue.

@M1chaelM , could you double-check that you get the same behavior? If not, could you remove your `build`, `install` and `log` directories and recompile to see if that makes any difference?

M1chaelM:
Confirmed this is working as expected for all tasks. The WAM-V unlocks for OnReady. It's up to the user to avoid driving through the navigation gate before the OnRunning state. To make things a little easier, I will adjust the default course for Navigation and Gymkhana tasks so it is a little further away from the WAM-V starting position.

M1chaelM:
Closing because this was not a bug and #554 makes it less likely that a new user will accidentally cross a gate before the plugin is in the "running" state.

PR #554, opened 2022-12-09
Move navigation course farther from WAMV

M1chaelM:
A very minor tweak to make the starting position of the navigation course a little farther away from the starting position of the WAMV in the navigation and gymkhana tasks. The purpose of this is to make it a little harder to accidentally cross the first gate before the simulation has entered the `OnRunning` state (see discussion at #545).

To test:
* Start the navigation task
  ```
  ros2 launch vrx_gz competition.launch.py world:=navigation_task
  ```
* Start teleop in a separate terminal.
  ```
  ```
* Try to drive the WAMV through the course as soon as possible.
* Confirm that the task enters the OnRunning state before you reach the first gate.
* Repeat for the gymkhana task.

### DIFF
--- a/vrx_gz/worlds/gymkhana_task.sdf
+++ b/vrx_gz/worlds/gymkhana_task.sdf
@@ -361,7 +361,7 @@
     <!-- The VRX short navigation course -->
     <include>
       <uri>model://short_navigation_course_0</uri>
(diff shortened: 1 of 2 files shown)


caguero (approved):
Works for me as expected!

PR #555, opened 2022-12-09
[Garden] Add PBR to the CPU cases

caguero:
This is the first pull request of a series for adding PBR to some of the models. In this case we start with the two CPU cases on the top plate of the WAM-V. Courtesy of Althone Torregosa.

Without PBR:
Screenshot from 2022-12-09 23-10-17

With PBR:

Screenshot from 2022-12-09 23-07-23


M1chaelM (approved):
Looks good!

PR #557, opened 2022-12-21
[Garden] Update environmental conditions

caguero:
This pull request simplifies the process of updating the environmental conditions. Right now, all the wave parameters need to match among all the plugins that accept a `<wavefield>` parameter to show a consistent behavior.

Additionally, this patch enables the wind for the WAM-V, as it wasn't working.

### How to test it?

Run the script to enable wind and change the waves:

```
cd <YOUR_VRX_WS>/src/vrx/vrx_gz/scripts
./update_sea_state_params.sh 1 5.0 "20000 0 0"
```

Verify the changes with:
```
git diff
```

Recompile:
```
cd <YOUR_VRX_WS>
```

Run the simulation:
```
```

The WAM-V should slowly move (wind) and the waves should be bigger.

### DIFF
--- a/vrx_gz/models/coast_waves/model.sdf
+++ b/vrx_gz/models/coast_waves/model.sdf
@@ -28,7 +28,7 @@
             <cell_count>300 300</cell_count>
             <wave>
               <model>PMS</model>
-              <period>5</period>
+              <period>5.0</period>
               <number>3</number>
               <scale>1.1</scale>
               <gain>0.3</gain>
--- a/vrx_gz/scripts/update_sea_state_params.sh
+++ b/vrx_gz/scripts/update_sea_state_params.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+# Copyright 2022 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


M1chaelM:
@caguero This script works for me, but it makes me a little nervous because it uses `find` and `sed` to change everything in the `worlds` and `models` directory to whatever values are specified. That saves us some hassle in a lot of situations, but if someone were to run it on the practice worlds where the values are intentionally different, it would make them all uniform and they would need to be restored manually or by using version control.

This isn't necessarily a big problem as long as we get the documentation right, but I wonder if it might be worth sketching out a longer-range plan to address this problem of having too many places where these environmental values are stored. We could either do that in this PR or in a separate issue if that's preferable. What do you think?

M1chaelM:
I created issue #564 to document what I see as the larger conceptual problem.

caguero:
> I created issue #564 to document what I see as the larger conceptual problem.



> @caguero This script works for me, but it makes me a little nervous because it uses `find` and `sed` to change everything in the `worlds` and `models` directory to whatever values are specified. That saves us some hassle in a lot of situations, but if someone were to run it on the practice worlds where the values are intentionally different, it would make them all uniform and they would need to be restored manually or by using version control.
> 
> This isn't necessarily a big problem as long as we get the documentation right, but I wonder if it might be worth sketching out a longer-range plan to address this problem of having too many places where these environmental values are stored. We could either do that in this PR or in a separate issue if that's preferable. What do you think?

That's a good observation. I tweaked the script for not touching subdirectories inside `world`. What do you think?

M1chaelM:
> That's a good observation. I tweaked the script ([5edc795](https://github.com/osrf/vrx/pull/557/commits/5edc79512e109d25f6c3c08a45c16df65b44efb1)) for not touching subdirectories inside `world`. What do you think?

This seems like a useful precaution. I think we can move forward with this for now with the idea that it may eventually be obsoleted by the resolution of #564

M1chaelM (approved):
Thanks for making this adjustment!

PR #558, opened 2022-12-22
[Garden] Fixing missing import

caguero:
This pull request fixes a Python issue when customizing a WAM-V. When the yaml configuration isn't compliant, instead of showing the expected error, we were showing the following error:

```
[generate_wamv.py-1] NameError: name 'yaml' is not defined
```

This patch fixes the problem.

### How to test it?

Follow [this tutorial](https://github.com/osrf/vrx/wiki/2023-Customizing-the-WAM-V) and you shouldn't get any unexpected error.

### DIFF
--- a/vrx_urdf/vrx_gazebo/src/vrx_gazebo/compliance.py
+++ b/vrx_urdf/vrx_gazebo/src/vrx_gazebo/compliance.py
@@ -4,6 +4,7 @@ from ament_index_python.packages import get_package_share_directory
 
 import numpy as np
 import os
+import rclpy
 import yaml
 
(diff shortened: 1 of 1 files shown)


M1chaelM:
When I attempt to run through the part of the tutorial that demonstrates behavior for a non-compliant wam-v, I get the following message instead:
```
NameError: name 'rclpy' is not defined
```
This corresponds with the actual fix made to `compliance.py`, so maybe 'yaml' was a typo?

M1chaelM (approved):
It seems clear that `yaml` was a typo, and this fix resolves the error that I would otherwise get when trying to generate a non-compliant wam-v.

PR #560, opened 2022-12-23
[Garden] Add PBR to the batteries

caguero:
Continuing the series of PRs for adding PBR to some of the models. In this case we modify the batteries on each pontoon. Courtesy of Althone Torregosa.

Wihout PBR:
Screenshot from 2022-12-23 15-46-51

With PBR:
Screenshot from 2022-12-23 15-53-02


M1chaelM (approved):
Looks good!

PR #561, opened 2022-12-23
[Garden] Add PBR to the camera and lidar

caguero:
Continuing the series of PRs for adding PBR to some of the models. In this case we modify the camera and laser visuals. Courtesy of Althone Torregosa.

Wihout PBR:
Screenshot from 2022-12-23 16-24-13

With PBR:
Screenshot from 2022-12-23 18-02-01


M1chaelM (approved):
This one was the most subtle, but I can see that with PBR the cameras are a bit darker and the surfaces reflect light so I think it's good to go.

Issue #563, opened 2022-12-29
WAM-V base link is far away from the WAM-V


M1chaelM:
Highlighting the WAM-V in Gazebo Sim shows a very large rectangle around the entity:

![wamv_box](https://user-images.githubusercontent.com/8611855/209887353-d4bdae8c-c3d3-4d11-9df4-adcb32faeecc.png)

Zooming in to the far end of the box reveals the reason is that the wam-v base link is located on the other side of the arena:

![wamv_base_link](https://user-images.githubusercontent.com/8611855/209887425-96fa1154-22dc-410f-891b-aff8c91edc95.png)

PR #566, opened 2023-01-04
[Garden] Fix issue #563 and inertias

caguero:
Fix issue #563.

The problem was in the `wamv_pinger.xacro`. We were creating a visual with an offset equals to the `position` of the pinger in world coordinates. This is wrong as the visual (if any) should represent the acoustic sensor and it should be inside the wam-v. 

I removed the visual and its collision as it's not very useful.

Bonus: I added some missing inertias, otherwise the default inertia values were incorrect. The inertias look good now:

Screenshot from 2023-01-04 17-20-04

### How to test it?

* Recheck the instructions described in #531 to confirm that the pinger is still working correctly.
* Click on the WAM-V. Verify that the bounding box is now correct.
* Right click and select `View->Inertia` to confirm that the inertias are reasonable. **Note that the inertia visualization is very slow the first time that you request it. This is a Gazebo issue.**

### DIFF
--- a/vrx_urdf/wamv_gazebo/urdf/components/wamv_pinger.xacro
+++ b/vrx_urdf/wamv_gazebo/urdf/components/wamv_pinger.xacro
@@ -4,19 +4,13 @@
     <link name="${namespace}/${name}">
       <visual>
         <origin xyz="0 0 0" rpy="0 0 0" />
-        <geometry>
-          <box size="0.1 0.1 0.1" />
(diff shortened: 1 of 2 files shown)


M1chaelM (approved):
* [x] The bounding box looks good. 
* [x] Inertia looks good (and makes a big difference in the behavior of the wam-v!). 
* [x] The pinger appears to still be working as expected

PR #572, opened 2023-01-05
Add missing texture

caguero:
We updated the WAM-V textures in #568 but the thrusters and propellers were still using the old texture. If you don't see this issue it's probably because you still have the `WamV.png` texture under your `install` directory. If you remove it from `install` (as we removed it from the code base), the thrusters and propellers are rendered in white material.

Ideally, the new texture should wrap the thruster and propeller but it's not 100% correct, so let's use it the old texture for now.


M1chaelM (approved):
Looks good. I deleted the install directory and rebuilt to verify. As a bonus I fixed the workflow in ci.yml

PR #590, opened 2023-02-10
Gymkhana practice worlds

caguero:
### How to test it?

Follow the instructions detailed in #540.  

Keep in mind that the pinger is in these locations for each practice world:
```
-503 302.5 0
-495 265 0
-533 293 0
```


M1chaelM:
After completing the navigation channel, the following message appears as expected:
```
[ruby $(which gz) sim-1] [Err] [ScoringPlugin.cc:359] VRX_EXIT_ON_COMPLETION and <per_plugin_exit_on_completion> both not set, will not shutdown on ScoringPlugin::Exit()
```
However, instead of just displaying once, the message displays repeatedly for the rest of the task. This effectively disables the console.

caguero:
> 505cab3

It should happen only once now. I modified the code in [505cab3](https://github.com/osrf/vrx/pull/590/commits/505cab3d40cd3cbd29952d5ed3851ba2efb9748e).

caguero:
Closing this one as it won't be available for VRX 2023 as we know it.

PR #597, opened 2023-02-20
Restore marker support

### LINKED ISSUE #596: stationkeeping / wayfinding visualization markers are not visible
When running stationkeeping and wayfinding practice and example tasks, it would be nice to have the visual markers for the goal / waypoints visible by default. Currently, they are not visible, and this functionality might not yet be ported from VRX classic.

caguero:
Fixes #596.

This PR restores the support to overlay a 3D marker on the scene. 

**Known limitations:** 
* Currently, it's not possible to render a `TEXT` marker. See [this open issue](https://github.com/gazebosim/gz-rendering/issues/487).
* It's not possible to pass a _text_ material like in Gazebo classic (`Gazebo/Red`). Instead, it's possible to set ambient and diffuse values. I decided not to expose it as it seems that we're always using the same color for the markers. Let me know otherwise and we could expose it.

### How to test it?

Launch one of the stationkeeping/wayfinding world and check that a white marker shows up. Example:

```
ros2 launch vrx_gz competition.launch.py world:=practice_2022_stationkeeping2_task
```

Screenshot from 2023-02-20 20-13-55

### DIFF
--- a/vrx_gz/CMakeLists.txt
+++ b/vrx_gz/CMakeLists.txt
@@ -81,6 +81,21 @@ install(
   TARGETS ScoringPlugin
   DESTINATION lib)
 
+# Stationkeeping scoring plugin
+add_library(StationkeepingScoringPlugin SHARED
+  src/StationkeepingScoringPlugin.cc
+  src/WaypointMarkers
+)
+target_link_libraries(StationkeepingScoringPlugin PUBLIC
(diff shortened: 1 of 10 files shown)


M1chaelM (approved):
Thanks! This is great. The color definitely does not matter. The text was sometimes useful for debugging new course designs for wayfinding, but I think we can live without it.

PR #598, opened 2023-02-27
Use a global synchronized wavefield

### LINKED ISSUE #564: Wavefield parameters are set in too many places
In our Gazebo Sim implementation, wavefield values must be specified separately for every instance PolyhedraBuoyancyDrag plugin, which must be attached to every buoyant object. Conceptually, this is bad because there is only one actual wavefield per body of water, and we never want to specify different wavefield values for different objects floating in the same body of water. Practically, it's bad because it means we have to manually keep all these values in sync, or use a script like the one introduced in PR #557 .

To fix the conceptual problem, we would ideally only have a single wave field associated with a body of water, and have our buoyancy plugin somehow access it. I'm not sure whether this is feasible right now.

A compromise we've used in the past is to use xacro variables to generate sdfs with appropriately synchronized values at build time. This has not been ported to GazeboSim and there may now be better tools for doing this in a clean way (e.g., maybe using embedded ruby or yaml would help clean this up?).

caguero:
Fixes #564 using Gazebo Transport.

Before this patch, all the plugins using a wave field needed to have the same `<wavefield>` SDF parameter block in its SDF plugin definition. This pull request changes this behavior by:

* Setting one of the plugins to operate as the `primary`. The `primary` plugin requires to specify all the `<wavefield>` parameters as usual. Then, it should periodically publish the wavefield parameters via Transport.
* The rest of the plugins should be set to operate as `secondary`. This means passing only one parameters inside the `<wavefield>` block. This parameters is the topic name where the wave field parameters are sent. When received, the wavefield can be populated with a new helper function.

Our `WaveVisual` plugin is set to operate as the primary. All the instances of `Surface` and `PolyhedreaBuoyancyDrag` plugins are set to operate as secondary.

### How to test it?

* You can run the simulation as usual:
```
ros2 launch vrx_gz competition.launch.py world:=stationkeeping_task
```

and verify that the buoys and the WAM-V move a bit up and down.

* You can modify `vrx_gz/worlds/stationkeeping_task.sdf`:
```
    <!-- The wave field -->
    <plugin filename="libPublisherPlugin.so" name="vrx::PublisherPlugin">
      <message type="gz.msgs.Param" topic="/vrx/wavefield/parameters"
               every="2.0">
...
        params {
          key: "gain"
          value {
            type: DOUBLE
            double_value: 0.1
          }
        }
...
      </message>
    </plugin>
```

Recompile and launch the simulation again. You should verify now that the WAM-V and the buoys stay flat at the water surface.
[Edit: updated test instructions]


caguero:
@M1chaelM , as we discussed offline, I've created a `PublishPlugin` to publish the wave field parameters every `2` seconds. All worlds in this PR use this plugin now.

M1chaelM (approved):

PR #609, opened 2023-03-24
Acoustic perception task

caguero:
The acoustic perception task including:
  * Scoring plugin
  * Example world
  * Practice worlds

How to test it?

1. Launch the simulation:
```
ros2 launch vrx_gz competition.launch.py world:=acoustic_perception_task
```

```
```

3. Echo the output of the acoustic sensor:
```
/wamv/pingers/pinger/range_bearing
```

Take a look at the range value and teleoperate the robot until you minimize the distance. When the range is <= 3 meters the simulation should finish.

Bonus: Originally I was planning to use the exit and entrance gate. We're not going to use it during the competition but I decided to include it in case it's useful for anyone. 

Screenshot from 2023-03-24 21-31-06


caguero:
> Everything seems to be working now but the pinger positions are so close together that it's hard to tell they have changed from world to world. Could we make the change a bit more obvious?

I just updated the pinger positions in the practice worlds to make them more different.

M1chaelM:
@caguero Your changes did make it easier to distinguish between the 3 worlds but I changed them even more to make it harder for teams to ignore the acoustic component and solve the task by driving around in a lawnmower pattern. (They can always do this, of course, but now it is more costly.)

PR #610, opened 2023-03-24
Follow-the-path task

caguero:
The follow the path task including:

* Example world
* Practice worlds
* We can reuse the navigation scoring plugin

How to test it?

1. Launch the simulation:
```
ros2 launch vrx_gz competition.launch.py world:=follow_path_task
```

```
```

3. Teleoperate the robot through the course. The task should successfully finish after crossing the last gate. Crossing the gates in the wrong error should also trigger the task to finish (unsuccessfully).

### DIFF
--- a/vrx_gz/models/short_navigation_course_0/model.sdf
+++ b/vrx_gz/models/short_navigation_course_0/model.sdf
@@ -171,6 +171,33 @@
       </plugin>
 
     </include>
+
+    <!-- Obstacles -->
+    <include>
+      <name>obstacle_0</name>
+      <pose>-20 -18 0.0 0 0 0 </pose>
+      <uri>https://fuel.gazebosim.org/1.0/openrobotics/models/mb_round_buoy_black</uri>
+      <plugin name="vrx::PolyhedraBuoyancyDrag"
+              filename="libPolyhedraBuoyancyDrag.so">
+        <fluid_density>1000</fluid_density>
+        <fluid_level>0.0</fluid_level>
+        <linear_drag>25.0</linear_drag>
+        <angular_drag>2.0</angular_drag>
+        <buoyancy name="collision_outer">
+          <link_name>link</link_name>
+          <pose>0 0 0 0 0 0</pose>
+          <geometry>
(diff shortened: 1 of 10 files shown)


M1chaelM:
I'm getting expected behavior from `follow_path_task` and `practice_2022_follow_path0_task`, which both use the same `short_navigation_course_0`. When I try to run `practice_2022_follow_path1_task` and `practice_2022_follow_path2_task`, I don't get any notification when I pass through gates and I'm unable to complete the task.

M1chaelM (changes requested):
This looks like it's working but I think we should add more gates for runs 1 and 2. The reason is that we're reverting to a task more like the one we had in VRX 2019, which specified 2-10 gates, rather than the 2-5 for the gymkhana navigation channel. To set expectations correctly, I think it would be good to have 3, 6, and 9-10 gates for practice worlds 0, 1, and 2, respectively.

Also, in world 2 I notice the buoys in the second gate are very far apart. This may be fine if they are within 25 meters, but I thought it might be worth double checking.

caguero:
> This looks like it's working but I think we should add more gates for runs 1 and 2. The reason is that we're reverting to a task more like the one we had in VRX 2019, which specified 2-10 gates, rather than the 2-5 for the gymkhana navigation channel. To set expectations correctly, I think it would be good to have 3, 6, and 9-10 gates for practice worlds 0, 1, and 2, respectively.
> 
> Also, in world 2 I notice the buoys in the second gate are very far apart. This may be fine if they are within 25 meters, but I thought it might be worth double checking.

All right, I just modified the navigation courses to have `3, `6` and `9` gates. I also added extra obstacles.

M1chaelM:
Courses 1 and 2 don't score past the third gate. I think because the scoring plugin initialization hasn't been updated to reflect the new models.

M1chaelM:
OK, I added those gates and it looks like it's working now.

M1chaelM (approved):

caguero:
> OK, I added those gates and it looks like it's working now.

Thanks!

PR #614, opened 2023-04-04
Acoustic wayfinding task

caguero:
This patch adds the new acoustic wayfinding task including:

* Example world
* Practice worlds
* Scoring plugin

How to test it?

1. Launch the simulation:
```
ros2 launch vrx_gz competition.launch.py world:=acoustic_wayfinding_task
```

```
```

3. Teleoperate the robot going close to each pinger until the task finishes.

### DIFF
--- a/vrx_gz/CMakeLists.txt
+++ b/vrx_gz/CMakeLists.txt
@@ -96,6 +96,21 @@ install(
   TARGETS WayfindingScoringPlugin
   DESTINATION lib)
 
+# AcousticWayfinding scoring plugin
+add_library(AcousticWayfindingScoringPlugin SHARED
+  src/AcousticWayfindingScoringPlugin.cc
+  src/WaypointMarkers
+)
+target_link_libraries(AcousticWayfindingScoringPlugin PUBLIC
+  gz-common${GZ_COMMON_VER}::gz-common${GZ_COMMON_VER}
+  gz-sim${GZ_SIM_VER}::core
+  gz-math${GZ_MATH_VER}
+  ScoringPlugin
+)
+install(
+  TARGETS AcousticWayfindingScoringPlugin
+  DESTINATION lib)
+
 # Plugins
 list(APPEND VRX_GZ_PLUGINS
   AcousticPingerPlugin
--- a/vrx_gz/src/AcousticWayfindingScoringPlugin.cc
+++ b/vrx_gz/src/AcousticWayfindingScoringPlugin.cc
@@ -1,0 +1,192 @@
+/*
+ * Copyright (C) 2023 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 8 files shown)


caguero:
As per our offline discussion this scoring plugin has been updated. It now tracks the error of the vehicle with respect to a single moving pinger. There are topics to track the instant error and the mean error during the task. I also introduced two obstacles that add a penalty when touched.

caguero:
> @caguero This is working as advertised, but I think we should consider a few tweaks if we have time before the release:
> 
> * I propose changing the name to "acoustic tracking" to avoid confusion with our previous wayfinding task and help suggest that we are practicing for working with an underwater vehicle. Open to other suggestions if you have a better name.
> * I think the underwater beacon should move at least twice as fast, and travel along a longer (and maybe slightly more complicated) path. Maybe something similar to path a USV would need to travel to complete the basic wayfinding task?
> * Can we re-use the obstacle field from the gymkhana task, or something similar to it? More obstacles would help give a better sense of what I'm hoping will be the final look of the task.
> 
> Let me know what you think. If you'd prefer to postpone any of the above until the next release I think that's fine too.

Good suggestions, thanks.

I renamed the plugin and changed the layout only in the `acoustic_tracking_task.sdf` world for now. Could you please test that world first before I actually modify the practice worlds? I didn't reuse the obstacle field model because I wanted to place the obstacle buoys in places that are close enough to the pinger trajectory.

M1chaelM:
> Could you please test that world first before I actually modify the practice worlds? I didn't reuse the obstacle field model because I wanted to place the obstacle buoys in places that are close enough to the pinger trajectory.

Thanks! I just tested it. I like this version much better. One thing I'm trying to brainstorm is ways to make it qualitatively different from the basic acoustic task. Right now it is different some of the time, but a lot of the time the beacon is moving through regions of open water and the USV just has to stationkeep above it (with a slight drift). I think we could address this easily with two changes:
* Make the beacon move even faster. Double its speed again? That way following it seems less like stationkeeping.
* Make sure there are situations where the USV has to decide between minimizing distance to the buoy and avoiding obstacles. Right now if I sit on top of the buoy I also avoid all the obstacles. Two ways to fix this would be to either change the path or just add a bunch more obstacles.

Also a tiny tweak, but could we change the color of the visual marker to something different from the obstacles? Maybe white or red?

M1chaelM:
> I implemented your suggestions bumping up the speed of the pinger, changing its visual to white and adding a few more obstacles in the trajectory. Let me know what you think.

Looks great! Thank you!

caguero:
> > I implemented your suggestions bumping up the speed of the pinger, changing its visual to white and adding a few more obstacles in the trajectory. Let me know what you think.
> 
> Looks great! Thank you!

@M1chaelM , I have updated the practice worlds. `#0` is very similar to the example world. In `#1` the pinger follows an `X` pattern over 4 waypoints. In `#2` the pinger follows a square pattern.

M1chaelM:
@caguero Per our offline discussion, could you adjust course 2 so it is more 2-dimensional, similar to course 1? Thank you!

caguero:
> @caguero Per our offline discussion, could you adjust course 2 so it is more 2-dimensional, similar to course 1? Thank you!

Done, I've reused the same waypoints as `#1` but to be followed in a different order and I shuffled the obstacles a bit to be in the trajectory of the pinger. For easier testing I left some orange buoys in the four waypoints that the pinger is visiting. I'll remove them before merging the PR.

M1chaelM (approved):
Looks good! Thanks!

PR #637, opened 2023-05-05
Jessica/wind plugin port

j-herman:
### To test:
```
ros2 launch vrx_gz competition.launch.py world:=practice_2022_stationkeepingX_task
```
You should see no movement due to wind in `stationkeeping0`, a slow drift towards the docks in `stationkeeping1`, and a faster drift towards the shore in `stationkeeping2`.

Wind speed and direction are published on `/vrx/debug/wind/speed` and `/vrx/debug/wind/direction`. As in gazebo classic, they're referenced to the Gazebo world frame. 

EDIT: Removed old description for the draft PR and copied testing instructions from below.


M1chaelM:
@j-herman Thanks! When it is ready for testing, could we swap out `test_wind_world` with the 3 versions of `practice_2023_stationkeeping#_task.sdf` configured to restore the original wind settings for those worlds (found [here](https://github.com/osrf/vrx/tree/gazebo_classic/vrx_gazebo/worlds/2022_practice))? I think this would help us get an intuitive sense of whether things are working as expected. Once it's set, we'll eventually update all the tasks with these settings.

j-herman:
[like]  Herman, Jessica (CIV) reacted to your message:
________________________________

j-herman:
> @j-herman Thanks! When it is ready for testing, could we swap out `test_wind_world` with the 3 versions of `practice_2023_stationkeeping#_task.sdf` configured to restore the original wind settings for those worlds (found [here](https://github.com/osrf/vrx/tree/gazebo_classic/vrx_gazebo/worlds/2022_practice))? I think this would help us get an intuitive sense of whether things are working as expected. Once it's set, we'll eventually update all the tasks with these settings.

Three stationkeeping 2022 practice worlds updated as requested.  They all run but I haven't verified the behavior matches expectations yet.

j-herman:
> @j-herman Thanks! When it is ready for testing, could we swap out `test_wind_world` with the 3 versions of `practice_2023_stationkeeping#_task.sdf` configured to restore the original wind settings for those worlds (found [here](https://github.com/osrf/vrx/tree/gazebo_classic/vrx_gazebo/worlds/2022_practice))? I think this would help us get an intuitive sense of whether things are working as expected. Once it's set, we'll eventually update all the tasks with these settings.

Three stationkeeping 2022 practice worlds updated as requested.  They all run but I haven't verified the behavior matches expectations yet.

j-herman:
There is a potential discrepancy between the legacy code implementation and the desired (published) behavior, unless I'm misunderstanding something.  In lines 254-260 of [usv_gazebo_wind_plugin.cc](https://github.com/osrf/vrx/blob/gazebo_classic/usv_gazebo_plugins/src/usv_gazebo_wind_plugin.cc) an apparent wind value is calculated (subtracting the vehicle's velocity from the wind vector), but the uncorrected wind is used in the force calculations.  I have included that correction in the new version.

Overall the ported plugin seems to be working as desired.  In comparison to the gazebo classic version, an equivalent force seems to produce less drift in the WAM-V.  I don't know whether other parameters may have changed in the model or in the port of the physics engines, or if there is something going on with how I am applying the force to the link.  There are a few more things I can troubleshoot in the implementation, but I do think this may be close enough to include in the next release even if we need to make tweaks later.  Qualitatively the behavior looks good and the interfaces are set.

Wind speed trace for stationkeeping1 world in gazebo classic:
image
Wind speed trace for stationkeeping1 world in gazebosim (note that the timescale is not in seconds - I haven't figured out how to fix that, so this graph represents more than 30 seconds):
image
Wind speed trace for stationkeeping2 world in gazebo classic:
image
Wind speed trace for stationkeeping2 world in gazebosim (same timescale issue):
image

In both worlds, the wind direction and the general behavior of the WAM-V were consistent - that is, the vehicle runs into the docks in stationkeeping1, and into the shore by the tent in stationkeeping2.

j-herman:
To test: 
`ros2 launch vrx_gz competition.launch.py world:=practice_2022_stationkeepingX_task`
You should see no movement due to wind in stationkeeping0, a slow drift towards the docks in stationkeeping1, and a faster drift towards the shore in stationkeeping2.  
Wind speed and direction are published on /vrx/debug/wind/speed and /vrx/debug/wind/direction.  As in gazebo classic, they're referenced to the Gazebo world frame.  A potential enhancement could be to also publish this information in the format that would be received from a weather station (ie, in knots and with the direction specified as the compass point from which the wind is blowing).

M1chaelM:
@j-herman Thanks! Should we mark this ready for review? If so I can test it.

j-herman:
Ah, I forgot about the ROS bridge!  I'll knock that out today.  The blank output is odd and I'll see what's causing that.

M1chaelM (approved):
Great. All looks good now. Thanks!

Issue #638, opened 2023-05-10
Problem with renaming model_name


Grigoriy0526:
Hello, I have a problem when starting the sydney_regatta world, if I rename model_name in the file competition.launch.py from 'wamv' to something else, an error occurs, what can such an error be related to?
![Снимок экрана от 2023-05-10 13-38-42](https://github.com/osrf/vrx/assets/133020151/3d2e6dbf-c637-4c5c-be9d-ed10e959c528)
![Снимок экрана от 2023-05-10 12-26-14](https://github.com/osrf/vrx/assets/133020151/fd3c4754-0f9a-40aa-82b7-573313237b66)

- Ubuntu 22.04
- ROS 2, Humble
- Gazebo Sim 7.0.0
-  VRX

carvee1:
I solved the bug yesterday, when spawning a vehicle with another name, the problem can be found in  the `vrx_urdf/wamv_gazebo/urdf/wamv_gazebo.urdf.xacro` file. In line 103, you have to change
 `<child>wamv/base_link</child>`
into 
 ` <child>$(arg namespace)/base_link</child>`
Afterwards it should work fine.
I already started the pull request #644  to solve this problem
Tested in ROS2 Humble and Ubuntu 22.04

M1chaelM:
Thanks @Grigoriy0526 and @carvee1 for finding and fixing this!
Resolved with #644

PR #644, opened 2023-05-16
Bug solution multi-vehicles in wamv_gazebo.urdf.xacro

carvee1:
When trying to launch a second vessel with the name wamv2, the following error message was shown 
`[ERROR] [launch]: Caught exception in launch (see debug for traceback): Error Code 20: Msg: Child frame with name[wamv/base_link] specified by joint with name[wamv_external_pivot_joint] not found in model with name[WAM-V]`
When inspecting the problem (also see problem #638), I found the problem in line 103 of the `wamv_gazebo.urdf.xacro` file, in which the namespace was not properly referenced to the namespace but only to the standard topic namespace of `wamv/base_link`. As a result without changing the child to `$(arg namespace)/base_link`, it is not possible to spawn a second vehicle with a different namespace. Therefore, without the change, it would not be possible to control the vessels independent of each other, as both subscribe to the same topic.

Tested the changes for ROS2


j-herman:
@M1chaelM @caguero - tested with locked and unlocked wamv, runs fine

M1chaelM:
@caguero Do you want to merge this bug fix into our upcoming release? @j-herman has reviewed, and the code change looks pretty straightforward, so I thought we might want to consider it even though we aren't focused on supporting multiple vehicles.

j-herman (approved):

PR #642, opened 2023-05-12
Updated wind plugin: practice worlds

j-herman:
Updating wind plugin for practice_2022 worlds

PR #649, opened 2023-05-17
Switching to the USVWind plugin in the rest of the worlds

M1chaelM:
In PR #642 we updated the worlds in `practice_2022` to use our custom wind plugin. This PR updates the rest of the example worlds included in the repo. Consistent with the wind settings in these example worlds, the wind velocity is set to 0, so the only effect should be to create the two wind speed and direction topics. The real purpose of doing this is to avoid having any confusing code snippets lying around.

To test:

`ros2 launch vrx_gz competition.launch.py world:=<taskname>_task`

In another terminal, confirm wind topics are publishing correctly through the ros bridge:
```
ros2 topic echo /vrx/debug/wind/direction
ros2 topic echo /vrx/debug/wind/speed
```


M1chaelM:
(Apologies, I seem to have accidentally pulled in some commits on another branch. I believe these have no effect since that branch has since been merged.)

M1chaelM:
@j-herman Do you have time to take a look? I was hoping to sneak this in before the release.

j-herman (approved):

Issue #640, opened 2023-05-12
`generate_wamv.launch.py` gives compliance error on empty WAMV


M1chaelM:
**Describe the bug**
Running `generate_wamv.launch.py` with empty configuration files produces an empty WAM-V platform as expected, but also gives the following error:

[generate_wamv.py-1] This component/thruster configuration is NOT compliant with the (current) VRX constraints. A urdf file will be created, but please note that the above errors must be fixed for this to be a valid configuration for the VRX competition.

**Expected behavior**
An empty WAM-V should be compliant. Are rules define maximum numbers of thrusters and components, but not minimums.

**To Reproduce**
List the steps to reproduce the problem:
1. Run 
```
touch empty_component_config.yaml
touch empty_thruster_config.yaml
ros2 launch vrx_gazebo generate_wamv.launch.py component_yaml:=`pwd`/empty_component_config.yaml thruster_yaml:=`pwd`/empty_thruster_config.yaml wamv_target:=`pwd`/wamv_target.urdf wamv_locked:=False
```
1. See error:
```
[generate_wamv.py-1] [ERROR] [1683911879.733853761] [configure_wamv]: 
[generate_wamv.py-1] This component/thruster configuration is NOT compliant with the (current) VRX constraints. A urdf file will be created, but please note that the above errors must be fixed for this to be a valid configuration for the VRX competition.
```

**System Configuration:**
Tell us about your system.
- OS: Ubuntu 22.04
- ROS Version: ROS 2, Humble
- Gazebo Version: Gazebo Sim 7.0.0
- Using VRX

M1chaelM:
### How this compliance check currently works:
1. `generate_wamv.launch.py` calls the script `generate_wamv.py` and passes it the required variables.
2. `generate_wamv.py` is a wrapper for `configure_wamv.py`.
3. `configure_wamv.py` uses the `create_component_xacro` and `create_thruster_xacro` files to generate the `urdf` and check for compliance at the same time.
4. These functions each call `create_xacro_file`, which is imported from `utils.py`
5. `create_xacro_file` takes two compliance check functions as arguments: one to check number of components/thrusters and one to check their parameters.
6.  The 4 compliance check functions that are passed to `create_xacro_file` are each imported from `compliance.py`, where they are defined as methods of a `ComponentCompliance` and `ThrusterCompliance` class. 
7.  If any of the 4 checks fail, the console will display the error listed above.
[...]

PR #658, opened 2023-05-23
`generate_wamv.py` should not produce a compliance error on an empty WAMV

M1chaelM:
This fixes issue #640

To test, follow this tutorial: https://github.com/osrf/vrx/wiki/empty_wamv_tutorial

Make sure there is no error message in the output; i.e. after running
```
ros2 launch vrx_gazebo generate_wamv.launch.py component_yaml:=`pwd`/empty_component_config.yaml thruster_yaml:=`pwd`/empty_thruster_config.yaml wamv_target:=`pwd`/wamv_target.urdf wamv_locked:=False
```
you should NOT see:
```
[generate_wamv.py-1] [ERROR] [1683911879.733853761] [configure_wamv]: 
[generate_wamv.py-1] This component/thruster configuration is NOT compliant with the (current) VRX constraints. A urdf file will be created, but please note that the above errors must be fixed for this to be a valid configuration for the VRX competition.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ jobs:
     runs-on: ubuntu-22.04
     container: 
       image: npslearninglab/watery_robots:humble_current
+      options: --user root
     steps:
       - run: sudo chown -R `whoami`:`whoami` .
       - name: Checkout vrx
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: Ubuntu CI
 on: [pull_request]
 
 jobs:
-  test-fortress-galactic:
+  test-humble-garden:
 
     runs-on: ubuntu-20.04
     container: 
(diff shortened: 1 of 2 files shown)


M1chaelM:
:confused: Testing is failing with the following error:
```
--- stderr: ros_gz_bridge
CMake Error at CMakeLists.txt:81 (find_package):
  By not providing "Findactuator_msgs.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "actuator_msgs", but CMake did not find one.

  Could not find a package configuration file provided by "actuator_msgs"
  with any of the following names:

    actuator_msgsConfig.cmake
    actuator_msgs-config.cmake

  Add the installation prefix of "actuator_msgs" to CMAKE_PREFIX_PATH or set
  "actuator_msgs_DIR" to a directory containing one of the above files.  If
  "actuator_msgs" provides a separate development package or SDK, be sure it
  has been installed.
```
I don't think this was produced by the 4 character change introduced in the PR. @caguero Are you aware of anything changing on the backend?

M1chaelM:
The recent change to `ros_gz` introduced a new dependency on `ros-humble-actuator-messages` (see the explanation here: https://github.com/gazebosim/ros_gz/issues/397). I've fixed this by adding it to the CI image. Long term, I think it would be better to move to a binary install of Gazebo Garden (see #662 ) since that's generally more accessible to our users.

M1chaelM:
@j-herman Do you have time to take a look at this? It should be pretty quick and it fixes the CI, which affects our other PRs.

j-herman (approved):
Works as expected - no errors

Issue #647, opened 2023-05-16
Rename 2022_practice directory and worlds


M1chaelM:
We ported the 2022 practice worlds from VRX Classic to VRX 2.0+ and put them in the [`vrx_gz/worlds/2022_practice/`](https://github.com/osrf/vrx/tree/main/vrx_gz/worlds/2022_practice) directory. While working on VRX 2023, we have updated these worlds and added to them so the name is increasingly misleading.

To resolve this we need to rename the `2022_practice` directory to `2023_practice`, and rename all the `sdf` files stored in the directory so they start with the prefix `practice_2023` instead of `practice_2022`. We then need to update the [phase 2](https://github.com/osrf/vrx/wiki/vrx_2023-phase2_dress_rehearsal) and [phase 3](https://github.com/osrf/vrx/wiki/vrx_2023-phase3_challenge) tutorials to refer to the 2023 practice worlds.

caguero:
I have updated the links to the practice worlds in these two tutorials.

M1chaelM:
This turned out to be a bit of a rabbit hole. There was no Phase 2 Task Practice tutorial for VRX 2023, so I created one from a copy of VRX 2022. However, the 2022 tutorial had links to a page for each task that described the features of each practice world (0, 1, and 2) for that task:

## Practice World Details
### [Task 1: Stationkeeping Practice Worlds](https://github.com/osrf/vrx/wiki/VRX-2022:-Task-1-Practice)
### [Task 2: Wayfinding Practice Worlds](https://github.com/osrf/vrx/wiki/VRX-2022:-Task-2-Practice)
### [Task 3: Perception Practice Worlds](https://github.com/osrf/vrx/wiki/VRX-2022:-Task-3-Practice)
### [Task 4: Wildlife Practice Worlds](https://github.com/osrf/vrx/wiki/VRX-2022:-Task-4-Practice)
### [Task 5: Gymkhana Practice Worlds](https://github.com/osrf/vrx/wiki/VRX-2022:-Task-5-Practice)
### [Task 6: Scan and Dock and DeliverPractice Worlds](https://github.com/osrf/vrx/wiki/VRX-2022:-Task-6-Practice)

[...]

PR #661, opened 2023-05-23
Rename 2022_practice worlds to 2023_practice

caguero:
See #647.

After recompiling you should be able to launch the worlds with:

```
ros2 launch vrx_gz competition.launch.py world:=practice_2023_follow_path1_task
```

Verify a few worlds to make sure that they load correctly.


M1chaelM (approved):
Looks good to me!

Issue #662, opened 2023-05-23
Change default Gazebo installation from source to binaries


M1chaelM:
We have been building Gazebo Sim from source, but it would simplify our installation process a bit if we could use the binaries instead. It would be great to make this switch for the final code freeze.

@caguero What do you think?

caguero:
I think we should be fine using binaries.

M1chaelM:
@caguero It looks like we can install Gazebo garden from binaries, but `ros_gz` is only available from source for Gazebo Garden. This is frustrating because it means our users have to build this code in the same workspace, which makes the process more complicated, makes the build take longer, and also prints lots of warning messages during the build process saying that the ignition namespace is deprecated.

Any ideas for improving this? Do you know why there is no binary for ros_gz garden?

caguero:
The binaries for this bridge combination are available but are a bit experimental, so we need to double-check that things are working. They're hosted in the prerelease Gazebo package repository.

https://packages.osrfoundation.org/gazebo/ubuntu-prerelease/pool/main/r/

Follow the `testing` instructions from: https://github.com/gazebo-tooling/release-tools/issues/895

bperseghetti:
@caguero @j-rivero  I think it would be great if we were also able to prerelease them for Iron ASAP as well. IE: `ros-iron-ros-gzgarden`

M1chaelM:
> The binaries for this bridge combination are available but are a bit experimental, so we need to double-check that things are working. They're hosted in the prerelease Gazebo package repository.

Amazing! Thank you! I thought we were out of luck. I'll give it a try and then work on getting the binary version into our CI image.

M1chaelM:
Note: this needs to happen before #643

M1chaelM:
OK, I've update the tutorial to use the `ros_gz` binaries. Installation is much nicer! I believe we can also build without setting GZ_VERSION every time, so I've removed this section:

> Note: Make sure to set the `GZ_VERSION` to `garden`. This is needed for building the `ros_gz` package.

It seems to build and run without that variable, but let me know if we need it for something I haven't understood. Similarly, I think we no longer need to set the pythonpath variable since we are not installing Gazebo Sim from source, so I am removing the following from the [Running VRX tutorial](https://github.com/osrf/vrx/wiki/running_vrx_tutorial):

> ## Set `PYTHONPATH` environment variable
> ```
> export PYTHONPATH=$PYTHONPATH:<YOUR_GAZEBO_WS>/install/lib/python
[...]

Issue #659, opened 2023-05-23
Upstream changes broke the `test-humble-garden` CI workflow


M1chaelM:
Upstream changes to `ros_gz` added a dependency that breaks our build. See discussion in #658. In addition, the github action checkout v2.4 has been deprecated due to deprecation of node12, so this needs to be upgraded to v3.0. Unfortunately, 3.0 does not support Dockerfiles that run as non-root users (and in general github actions do not support this), so it is incompatible with our test image.

M1chaelM:
CI has been restored by adding the missing dependency to the test image and by #658, which upgrades our workflow to use checkout v3, and overrides the user to run as root.

Two further changes would improve our CI:
* Until github actions supports running as a non-root user, the Docker image should not create a user, thereby eliminating the need to override it. I think we can instead create the user dynamically using github actions during the test process.
* We should either use rosdep or change our install process to use debs for `gazebo sim` and `ros_gz` so upstream changes don't break our build. Since we no longer need to build from source, I think switching to debs makes the most sense.

M1chaelM:
#663 switches over to using debs, so changes in dependencies should now be handled for us automatically. This issue is therefore hereby resolved.

PR #663, opened 2023-05-26
improving CI workflow

M1chaelM:
This PR partially addresses #659 by modifying our workflow to expect a Docker image that is running as root by default.


M1chaelM:
Note: before merging this I would like to resolve #662 and update our test image to match the new install process.

PR #664, opened 2023-05-27
files for rviz tutorial

j-herman:
This will add a launch file and RViz config file to work with the new tutorial.  
Work still to go: add a truth-data odometry publisher so the vehicle will move in RViz for testing and visualization purposes.  The convenient ways to do this in Gazebo classic are no longer available so I'll have to write something up.


j-herman:
To test: try the [tutorial](https://github.com/osrf/vrx/wiki/rviz_tutorial)
I made a custom wam-v for this tutorial as the work required to enable ground truth from the command line was getting to be excessive and isn't very useful.  If desired, I can add a note here on how to enable ground truth before creating a custom wam-v, and what to change in the rviz launch file to load that vehicle - that is much easier.

M1chaelM:
> Documentation for RViz2 has not been updated on the wiki, so for additional details you may be interested in looking at the [codebase](https://github.com/ros2/rviz/tree/humble)
:laughing:

M1chaelM:
Looks like this uncovers another missing dependency we should add to the installation instructions:
```
username@domain:~/vrx_ign$ ros2 launch vrx_gazebo rviz.launch.py
[INFO] [launch]: All log files can be found below /home/username/.ros/log/2023-06-05-15-27-51-986336-domain-2666437
[INFO] [launch]: Default logging verbosity is set to INFO
[ERROR] [launch]: Caught exception in launch (see debug for traceback): "package 'joint_state_publisher' not found, searching: ['/home/username/vrx_ign/install', '/opt/ros/humble']"
```
Resolved by:
```
sudo apt install ros-humble-joint-state-publisher
```

M1chaelM:
@j-herman Thanks for putting this together. It seems to be working for me, but I'm getting one unexpected behavior, which is that if I teleop the WAM-V it drives off the Rviz display. Previously, my memory is that the Rviz display remained centered on the WAM-V.

* Do you think I'm remembering that correctly?
* If so, what can we do to restore that behavior or something like it?

j-herman:
> @j-herman Thanks for putting this together. It seems to be working for me, but I'm getting one unexpected behavior, which is that if I teleop the WAM-V it drives off the Rviz display. Previously, my memory is that the Rviz display remained centered on the WAM-V.
> 
>     * Do you think I'm remembering that correctly?
> 
>     * If so, what can we do to restore that behavior or something like it?

That is an easy fix - if you prefer that behavior I will just change the fixed frame for the view.  I actually was playing with both options and we could leave them both in as a demo of that parameter.

j-herman:
@caguero Tagging for a quick look to see if we could pass some of the wamv arguments (specifically, ground_truth_enabled) from the command line when running the competition.launch script

j-herman:
* Updated [tutorial](https://github.com/osrf/vrx/wiki/rviz_tutorial)
* Revised launch file for RViz
* New node to publish the robot description for RViz to consume (previously used robot_state_publisher, but we should not publish the transforms again)
* Fixed a missing input to the P3D node to enable ground truth (no longer used here)

j-herman:
Tagging [this issue](https://github.com/osrf/vrx/issues/681) which captures the not-quite-ideal behavior in this demo.  Still better than not having one, I think.

M1chaelM:
I see the behavior you describe, but I'm on the fence about whether we should merge this. My biggest concern is that running the commands shown in the tutorial does not produce a display that looks like what we are showing it should look like in the tutorial. Maybe we could merge this and put a note on the tutorial saying it doesn't actually work? Or we could try to fix those transform frames. My guess is that they are the result of variable arguments for prefixes being inconsistently applied in our code--not great but I would hope not too hard to track down and resolve.

j-herman:
Closing - obsolete due to merging of [PR698](https://github.com/osrf/vrx/pull/698)

Issue #641, opened 2023-05-12
wamv pinger (sensor) is listed as a component but no longer has visual or collision


M1chaelM:
**Describe the bug**
Our `example_component_config.yaml` lists a pinger as a component of the WAMV, and our `wamv_gazebo.urdf.xacro` file suggests it is a component of the WAM-V (though it's default location is strange), but specifying its inclusion now has no effect because the collision and visual were removed in #566. Since the component is not actually required to receive messages from the broadcasting pinger, I'm not sure whether it should be mentioned or not. 

**Expected behavior**
Ideally, the pinger range_bearing topic would only be active if the WAM-V had an acoustic sensor enabled, an enabling it would cause it to show up as part of the WAMV's entity tree.

**To Reproduce**
List the steps to reproduce the problem:
1. Run 'ros2 launch vrx_gz competition.launch.py world:=sydney_regatta'
2. Look at the WAMV in the entity tree and notice there is no "pinger" listed.

caguero:
Fixed in #669

PR #669, opened 2023-06-06
Fix issue #641

caguero:
See issue #641.

The problem was our well known Gazebo "feature" that eliminates fixed joints when converting from `URDF` to `SDF`. Then, the pinger link disappeared. I converted the fixed joint into a revolute joint with no movement to bypass the problem.

### How to test it?

* Launch the simulation as usual:
```
```

* Look at the WAMV in the entity tree and verify that there is a "pinger" listed. If you select the pinger link, you should observe a tiny sphere between the two pontoons as I enabled a small visual for reference.

* Additionally, if you comment out the line `221` in `wamv_gazebo.urdf.xacro`:
```
<xacro:wamv_pinger name="pinger" position="-528 191 -2.0" />
```
You shouldn't see the pinger link in the WAMV entity tree and the functionality of the pinger won't be available (`gz topic -l` shouldn't report the topic `/wamv/pingers/pinger/range_bearing`).

### DIFF
--- a/vrx_gz/src/vrx_gz/bridges.py
+++ b/vrx_gz/src/vrx_gz/bridges.py
@@ -46,22 +46,6 @@ def cmd_vel(model_name):
         ros_type='geometry_msgs/msg/Twist',
         direction=BridgeDirection.ROS_TO_GZ)
 
-def acoustic_pinger(model_name):
-    return Bridge(
-        gz_topic=f'{model_name}/pingers/pinger/range_bearing',
-        ros_topic=f'pingers/pinger/range_bearing',
-        gz_type='ignition.msgs.Param',
-        ros_type='ros_gz_interfaces/msg/ParamVec',
-        direction=BridgeDirection.GZ_TO_ROS)
-
-def set_acoustic_pinger(model_name):
-    return Bridge(
-        gz_topic=f'{model_name}/pingers/pinger/set_pinger_position',
-        ros_topic=f'pingers/pinger/set_pinger_position',
-        gz_type='ignition.msgs.Vector3d',
-        ros_type='geometry_msgs/msg/Vector3',
-        direction=BridgeDirection.ROS_TO_GZ)
-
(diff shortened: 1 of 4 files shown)


M1chaelM:
@caguero This seems to be working but there is a little quirk where even though `gz topic -l` doesn't show `/wamv/pingers/ping/range_bearing`, `ros2 topic list` still includes it:
```
/clock
/parameter_events
/rosout
/tf
/tf_static
/vrx/contacts
/vrx/debug/wind/direction
/vrx/debug/wind/speed
/vrx/task/info
/wamv/pingers/pinger/range_bearing
/wamv/pingers/pinger/set_pinger_position
/wamv/pose
...
```
If I try to echo it I get nothing, so maybe it's ok?

caguero:
> @caguero This seems to be working but there is a little quirk where even though `gz topic -l` doesn't show `/wamv/pingers/ping/range_bearing`, `ros2 topic list` still includes it:
> 
> ```
> /clock
> /parameter_events
> /rosout
> /tf
> /tf_static
> /vrx/contacts
> /vrx/debug/wind/direction
> /vrx/debug/wind/speed
> /vrx/task/info
> /wamv/pingers/pinger/range_bearing
> /wamv/pingers/pinger/set_pinger_position
> /wamv/pose
> ...
> ```
> 
> If I try to echo it I get nothing, so maybe it's ok?

I just tweaked the code to not advertise the ROS 2 topic if the pinger isn't present.

M1chaelM (approved):
With this last change I'm seeing that the `range_bearing` topic is no longer advertised on the ROS side, but also that `set_pinger_position` is not advertised on either side (i.e. nothing related to the pinger shows up in either `gz topic -l` or `ros2 topic list`). This seems good to me for now. If we ever extend to multiple vehicles we may want to revisit (and maybe figure out a way to differentiate the transmitting pinger from the receiving pinger).

Issue #650, opened 2023-05-17
Set Waves to a Non-zero steepness


M1chaelM:
Our wave model uses trochoidal waves but we currently have the steepness set to zero in all worlds, which collapses the shape to a sinusoid. If possible we should update this value to something that makes the trochoidal shape a little more evident.

hashoak:
#673

M1chaelM:
Per discussion in #674 it seems that updating the steepness does not have any discernible effect on the wave rendering, and no one is sure why this is the case. Consequently, we don't need to update the values right now, and I will close this issue. Since this situation isn't urgent but also not ideal, I'll make a new issue to document the problem.

PR #673, opened 2023-06-09
#650 issue: updated all the steepness values to 1

hashoak:
Have changed all the steepness values to 1 as said in the comments for the variable `steepness` => "Control the wave steepness. 0 is sine waves, 1 is Gerstner waves."

### DIFF
--- a/vrx_gz/models/short_navigation_course_1/model.sdf
+++ b/vrx_gz/models/short_navigation_course_1/model.sdf
@@ -38,7 +38,7 @@
             <angle>0.4</angle>
             <tau>2.0</tau>
             <amplitude>0.0</amplitude>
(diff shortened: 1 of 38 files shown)


M1chaelM:
Thanks for this contribution! My apologies, but I'm afraid the issue you addressed was a bit underspecified--we were discussing it internally and I didn't expect a community member to take it on so quickly. In attempting to test it, we discovered another issue with steepness that is causing us to postpone changes to this value for now. I've written this up in issue #683 (which I hope is more clear than what I wrote for #650).If you're interested in digging into this, we'd certainly welcome help investigating!

In the meantime I will close this PR since I don't anticipate wanting to set all steepness values to 1.

Issue #665, opened 2023-05-31
Wave Surface stays flat


alexdrc00:
I was looking to run some tests with waves of greater magnitudes and raised peak period and gain to higher values, 5 and 8 for example, and although I could see the wamv clearly affected by these changes, oscillating up and down, I could not see any changes in the water's visuals, just a flat surface, conversely to classic where I can see the visuals changing. I assume this has not been ported yet but wanted to ask if it is the expected behavior just in case.

M1chaelM:
This is not expected. Could you please fill out a bug report? See below:

**Describe the bug**
Please say what is broken. Include whether you think this is a problem with the code or the documentation / tutorials.

**Expected behavior**
What you expected to happen.

**To Reproduce**
List the steps to reproduce the problem:
1. Run '...'
2. Next, run '....'
3. See error

**System Configuration:**
Tell us about your system.
- OS: [e.g. Ubuntu 22.04]
- ROS Version: [e.g. ROS 2, Humble]
- Gazebo Version: [e.g. Gazebo Sim 7.0.0]
- Graphics Card: [e.g. Nvidia GTX 650]
- Are you using VRX or VRX Classic? (The default is VRX).

**Screenshots**
If applicable, add screenshots to help explain your problem. (If the problem is terminal output, it is better to cut and paste the text than to take a screenshot.)

**Additional notes**
Add any other context about the problem here.

M1chaelM:
Is this running in a Docker container on WSL2? Are you seeing a very low RTF? Since the wave visual is generated by the graphics card I think the problem may be that you aren't using a separate Nvidia graphics card. The sky in the screenshot you sent does not look quite right either.

alexdrc00:
No, this is runing on an Linux Device. As for the RTF, it's not too bad, I see around 0.8 rtf running the whole sydney_regatta world.

M1chaelM:
OK, that RTF is about right. I'm a little surprised. However, the system requirements for this do still include an Nvidia graphics card. If you can reproduce this issue on a system that meets the [system requirements](https://github.com/osrf/vrx/wiki/system_requirements) we can try to troubleshoot it with you. Otherwise, you may be able to get it to work but there won't be official support.

alexdrc00:
[...]

PR #674, opened 2023-06-10
update order of wavefield parameters message

### LINKED ISSUE #651: Improve Wavefield parameter tutorial and message
Our [theory of operations](https://github.com/osrf/vrx/wiki/vrx_theory) gives the following summary of the wave field parameters listed in our paper:

 * `period` ($T_p$, "peak period")
* `gain` ($K_H$, "wave height gain")
* `direction` ($\theta$, "mean wave direction")
* `steepness` ($q$, "wave steepness")

However, the message we use to set the wave field seems to use a vector to specify direction instead of an angle. We need to either change this or make a tutorial that explains it.

Also, if it's possible to remove the unused parameters in our wavefield message it would improve readability.

tejalbarnwal:
Aims to close #651 and #650

## Sub-Tasks
-  [x] Update the order inside the wavefield parameter message published with the `.sdf` file
-  [ ] Update the steepness parameter to have trochoidal shaped waves

## Test it
Launching one of the  2023 practice worlds with a considerable gain parameter(like say 0.8) would work
```
ros2 launch vrx_gz competition.launch.py world:=2023_practice/practice_2023_stationkeeping2_task
```

### DIFF
--- a/vrx_gz/worlds/2023_practice/practice_2023_acoustic_perception0_task.sdf
+++ b/vrx_gz/worlds/2023_practice/practice_2023_acoustic_perception0_task.sdf
@@ -480,6 +480,36 @@
       <message type="gz.msgs.Param" topic="/vrx/wavefield/parameters"
                every="2.0">
         params {
+          key: "direction"
+          value {
+            type: VECTOR3D
+            vector3d_value {
+              x: 1
+            }
+          }
+        }
+        params {
+          key: "gain"
+          value {
+            type: DOUBLE
+            double_value: 0.3
+          }
+        }
+        params {
+          key: "period"
+          value {
+            type: DOUBLE
+            double_value: 5
+          }
+        }
+        params {
+          key: "steepness"
+          value {
+            type: DOUBLE
+            double_value: 0
+          }
+        }
+        params {
           key: "amplitude"
           value {
             type: DOUBLE
@@ -504,22 +534,6 @@
           }
         }
         params {
-          key: "direction"
-          value {
-            type: VECTOR3D
-            vector3d_value {
-              x: 1
-            }
-          }
-        }
-        params {
-          key: "gain"
-          value {
-            type: DOUBLE
-            double_value: 0.3
-          }
-        }
-        params {
           key: "model"
           value {
             type: STRING
@@ -531,13 +545,6 @@
           value {
             type: INT32
             int_value: 3
-          }
-        }
-        params {
-          key: "period"
-          value {
-            type: DOUBLE
-            double_value: 5
           }
         }
         params {
@@ -562,13 +569,6 @@
               x: 6000
               y: 6000
             }
-          }
-        }
-        params {
-          key: "steepness"
-          value {
-            type: DOUBLE
-            double_value: 0
           }
         }
         params {
--- a/vrx_gz/worlds/2023_practice/practice_2023_acoustic_perception1_task.sdf
+++ b/vrx_gz/worlds/2023_practice/practice_2023_acoustic_perception1_task.sdf
@@ -480,6 +480,36 @@
       <message type="gz.msgs.Param" topic="/vrx/wavefield/parameters"
                every="2.0">
         params {
+          key: "direction"
+          value {
+            type: VECTOR3D
+            vector3d_value {
+              x: 1
+            }
+          }
+        }
+        params {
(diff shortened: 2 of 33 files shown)


tejalbarnwal:
Also, I found that the `2023_practice/practice_2023_follow_path0_task*.sdf` files don't include a publisher plugin to publish the wavefield parameters. I wanted to confirm if this is intentional or if I should add that snippet.

caguero:
> Also, I found that the `2023_practice/practice_2023_follow_path0_task*.sdf` files don't include a publisher plugin to publish the wavefield parameters. I wanted to confirm if this is intentional or if I should add that snippet.

That wasn't intentional if that's the case.

tejalbarnwal:
> > Also, I found that the `2023_practice/practice_2023_follow_path0_task*.sdf` files don't include a publisher plugin to publish the wavefield parameters. I wanted to confirm if this is intentional or if I should add that snippet.
> 
> That wasn't intentional if that's the case. Good catch!

Hey, I will update the respective world files.


I also wanted to share some experimental tuning that I performed. As Jessica mentioned in the previous weekly meeting, keeping the steepness parameter to 1 could have led to the formation of loops. 
I first tried keeping the steepness to 0.5 and then slowly increasing it to 1.0 but could not observe considerable changes.
I also tried increasing the value to 10.0, hoping that maybe it would help me observe, but I couldn't.
Therefore, I created some recordings with different values of steepness parameters. I have attached them [here](https://drive.google.com/drive/folders/1OjZ66nuT2jZUqqeIKWiYMpnEO-UtXOLZ?usp=sharing).
It is possible I might be missing something, and it would really help if you could provide your input.

j-herman:
Hi Tejal,
I took a look through the code for the wave field to try to refresh my memory on how this should work.  You won't see anything change with steepness parameters above 1 because the values get clipped.  
Some ideas to try for troubleshooting - zoom in a bit more than what you have in those video clips, and see if it helps distinguish wave characteristics.  Try varying the gain parameter, maybe with zero wind, to see if you are getting appropriate visuals to match the wave size.  You can look at the VRX Classis videos to get an idea of what the waves were like, in case that helps. 
I tried running a few world's on my laptop to look at this, and the surface visuals really didn't seem to be varying the way I would have expected.  Wondering if maybe we are seeing something like [issue 665](https://github.com/osrf/vrx/issues/665)?  I'll need to spend a little more solid time on this to be sure, and I want to run it in docker as well.

caguero:
> I experimented with disabling the wind plugin altogether and tried playing with gain, steepness, and even period as well. The effect of changing the period is evident, but that is not the case for steepness, even when I set the gain value to maximum(1.0). 

One important detail is that `gain` can go bigger than `1`.  Could you double-check if `steepness` is more evident with higher values of `gain`?

caguero:
> > I experimented with disabling the wind plugin altogether and tried playing with gain, steepness, and even period as well. The effect of changing the period is evident, but that is not the case for steepness, even when I set the gain value to maximum(1.0).
> 
> One important detail is that `gain` can go bigger than `1`. Could you double-check if `steepness` is more evident with higher values of `gain`?

Another idea could be to log some values of `P.x`, `P.y`, `P.z` from the shader (https://github.com/osrf/vrx/blob/main/vrx_gz/models/coast_waves/materials/programs/GerstnerWaves_vs_330.glsl#L120-L122) and plot them to see if we see something interesting changing the `steepness`.

tejalbarnwal:
> > > I experimented with disabling the wind plugin altogether and tried playing with gain, steepness, and even period as well. The effect of changing the period is evident, but that is not the case for steepness, even when I set the gain value to maximum(1.0).
> > 
> > 
> > One important detail is that `gain` can go bigger than `1`. Could you double-check if `steepness` is more evident with higher values of `gain`?
> 
> Another idea could be to log some values of `P.x`, `P.y`, `P.z` from the shader (https://github.com/osrf/vrx/blob/main/vrx_gz/models/coast_waves/materials/programs/GerstnerWaves_vs_330.glsl#L120-L122) and plot them to see if we see something interesting changing the `steepness`.

Okay, i will try it out today

tejalbarnwal:
As discussed in the meeting, we can table this discussion for another time, as it appears that the steepness of ocean waves does not have a visual impact. Therefore, closing this PR.

Issue #680, opened 2023-06-23
Many topics don't publish if world is specified with an absolute path


M1chaelM:
**Describe the bug**
When launching the VRX competition with an argument such as `world:=/full/path/to/stationkeeping_task` the world seems to launch but many topics are not published. On the ROS side, I get:
```
/clock
/parameter_events
/rosout
/tf
/tf_static
/vrx/debug/wind/direction
/vrx/debug/wind/speed
/vrx/task/info
/wamv/pose
/wamv/pose_static
/wamv/sensors/cameras/front_left_camera_sensor/optical/camera_info
/wamv/sensors/cameras/front_left_camera_sensor/optical/image_raw
/wamv/sensors/cameras/front_right_camera_sensor/optical/camera_info
/wamv/sensors/cameras/front_right_camera_sensor/optical/image_raw
/wamv/sensors/cameras/middle_right_camera_sensor/optical/camera_info
/wamv/sensors/cameras/middle_right_camera_sensor/optical/image_raw
```
This is a problem because to run our competition we rely on the ability to specify the path of world files that are not installed in the install directory.

**Expected behavior**
I expect to see at least the following topics:
```
/clock
/parameter_events
/rosout
/tf
/tf_static
/vrx/contacts
/vrx/debug/wind/direction
/vrx/debug/wind/speed
/vrx/stationkeeping/goal
/vrx/stationkeeping/mean_pose_error
/vrx/stationkeeping/pose_error
/vrx/task/info
/wamv/pingers/pinger/range_bearing
/wamv/pingers/pinger/set_pinger_position
/wamv/pose
/wamv/pose_static
/wamv/sensors/cameras/front_left_camera_sensor/camera_info
/wamv/sensors/cameras/front_left_camera_sensor/image_raw
/wamv/sensors/cameras/front_left_camera_sensor/optical/camera_info
/wamv/sensors/cameras/front_left_camera_sensor/optical/image_raw
/wamv/sensors/cameras/front_right_camera_sensor/camera_info
/wamv/sensors/cameras/front_right_camera_sensor/image_raw
/wamv/sensors/cameras/front_right_camera_sensor/optical/camera_info
/wamv/sensors/cameras/front_right_camera_sensor/optical/image_raw
/wamv/sensors/cameras/middle_right_camera_sensor/camera_info
/wamv/sensors/cameras/middle_right_camera_sensor/image_raw
/wamv/sensors/cameras/middle_right_camera_sensor/optical/camera_info
[...]

PR #687, opened 2023-06-27
add debug option to wind publishers

j-herman:
We will not want to publish the wind speed and direction during the actual competition events.  Some options for enabling/disabling these topics, in order of difficulty of implementation, are:
* add a <debug> tag to the wind plugin sdf, and set it to false when creating competition worlds (or, we could do the opposite and have it default to false)
* add an argument into launch.py that tells us whether we want to run in competition or debug mode, which adds topics to the bridge as needed
* change how we're selecting which topics to bridge/ publish, as described in the comments for [680](https://github.com/osrf/vrx/issues/680)

This PR implements the first option.  To test, add the <debug> tag to a world file, launch vrx with that world, and observe that no data is published to `/vrx/debug/wind/direction` or `/vrx/debug/wind/speed`.

For example:
* add `<debug>false</debug>` to stationkeeping_task.sdf around line 604
* run `ros2 launch vrx_gz competition.launch.py world:=stationkeeping_task`
* run `ros2 topic echo /vrx/debug/wind/direction`

### DIFF
--- a/vrx_gz/src/USVWind.cc
+++ b/vrx_gz/src/USVWind.cc
@@ -85,6 +85,10 @@ public:
 public:
   double lastPublishTime = 0;
 
+  /// \brief Debug option to turn off wind information publishers
+public:
+  bool debug = true;
+
   /// \brief Transport node
 public:
   transport::Node node;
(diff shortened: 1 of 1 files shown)


tejalbarnwal:
Hey, 
I tested it out. It works flawlessly!

tejalbarnwal (approved):

j-herman:
@tejalbarnwal Thank you for the review!  I'm going to hold this until we figure out the implementation for competition mode as we may want to change the method.

caguero:
> @caguero Just confirming we don't need this before I close it - fully covered by the competition mode PR?

Yeah, feel free to close it. We should be able to do it by isolating all Gazebo topics and only let the ROS 2 topics go through the competitors image.

PR #694, opened 2023-06-28
add command line argument for passing extra args to gz sim

M1chaelM:
This PR adds the `extra_gz_args` option to `competition.launch.py` and passes it through to `gz sim`. We had something similar in VRX classic so we could use gazebo options without having to implement command line arguments for all of them. (There might be a cleaner way to do this; if so please advise.)

The immediate reason for doing this is so we can use gz sim's recording functionality, which we need to get playback working.

### To test
Try passing some arguments through to gz sim:
```
ros2 launch vrx_gz competition.launch.py headless:=false world:=stationkeeping_task extra_gz_args:="--record_period 0.01 --record_path $HOME/.gz_sim_arg_test --log_overwrite"
```
Check that logs are written to `$HOME/.gz_sim_arg_test`.

Bonus: adding support for "paused" argument. To test:
```
ros2 launch vrx_gz competition.launch.py paused:=true world:=stationkeeping_task
```
The simulation should start in a paused state.

### DIFF
--- a/vrx_gz/launch/competition.launch.py
+++ b/vrx_gz/launch/competition.launch.py
@@ -32,7 +32,7 @@ def launch(context, *args, **kwargs):
     robot = LaunchConfiguration('robot').perform(context)
     headless = LaunchConfiguration('headless').perform(context).lower() == 'true'
     robot_urdf = LaunchConfiguration('urdf').perform(context)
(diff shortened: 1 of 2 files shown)


caguero (approved):
Works as expected.

Issue #681, opened 2023-06-23
TF frames are inconsistent 


j-herman:
Placeholder issue for some problems with how the Ros bridge is re-publishing the wamv TF messages.  I will add detail and a recommendation for correction.  This has become evident while working on the RVIZ tutorial and finishing that work will be dependent on closing this issue.
More specifically, we seem to have an extra "wamv/" inserted into some frames but not all, which breaks the TF tree when read from ROS.  Still early in troubleshooting, so this may not actually be the problem.

j-herman:
**Describe the bug**
The transforms published by gz::sim::systems::PosePublisher do not match the tree created in the WAM-V URDF.  Behavior is correct within Gazebo but this causes problems when visualizing the vehicle in RVIZ, and could be an issue for other consumers of TF2 data.
The major differences noted:
* PosePublisher prepends a namespace or parent frame name on each frame when publishing 
* PosePublisher is set up to publish sensor transforms, which are not part of the urdf
* PosePublisher does not publish any transforms for the sensor or battery links, which are part of the urdf
* The TF tree adds the wamv model (not part of the urdf) with all children descending from there.  The original tree structure of the urdf, with all links as children of wamv/base_link, is lost.  Since wamv and wamv/base_link are connected this may be functionally ok.
See [frames_2023-06-23_08.55.40.pdf](https://github.com/osrf/vrx/files/11871486/frames_2023-06-23_08.55.40.pdf) for a side-by-side comparison of the transforms from PosePublisher and the transforms produced by running robot_state_publisher on the same urdf.

**Expected behavior**
I expect the TF tree published by the simulation to match the URDF, both in frame names and in parent-child relationships.  

**To Reproduce**
Run VRX:
`ros2 launch vrx_gz competition.launch.py world:=stationkeeping_task`

Echo the gz or ros topics to see which transforms are published:
`ros2 topic echo /tf`
`ros2 topic echo /tf_static`
`gz topic --echo --topic /model/wamv/pose`
`gz topic --echo --topic /model/wamv/pose_static`

To generate the transform tree from the URDF, try:
`ros2 run robot_state_publisher robot_state_publisher ~/vrx_ws/install/share/vrx_gazebo/models/wamv/tmp/model.urdf'`

Visualize both sets of frames:
`ros2 run tf2_tools view_frames`
 
**System Configuration:**
Tell us about your system.
- Ubuntu 22.04
- Humble
- Gazebo Sim 7
- Nvidia RTX 3080
[...]

PR #698, opened 2023-06-28
move tf to ros

j-herman:
This PR is an alternative to PR [664](https://github.com/osrf/vrx/pull/664) which also resolves many of the TF issues in [681](https://github.com/osrf/vrx/issues/681).  It changes the way we publish TF information on the ROS side by:
* adding a JointStatePublisher plugin to the wamv and bridging the topic to ROS to make dynamic joint states available on the ROS side
* modifying the PosePublisher plugin to only publish sensor transforms, as those are only available via the sdf and will not be published when parsing the urdf
* replacing most of the functionality in the custom pose_tf_broadcaster node with a robot_state_publisher node, which provides the robot description and tf tree in a format that can be ingested by ROS applications such as RViz

A launch file to start an instance of RViz with a custom config file is included for testing and for the tutorial.

UNKNOWN, help needed: Is any other process on the Gazebo side using the transforms from the sdf that PosePublisher was producing previously?  I think not, as we have access to the truth data, but if so this PR will likely break something and needs to be reworked.

To test:

`ros2 launch vrx_gz competition.launch.py world:=stationkeeping_task`
`ros2 launch vrx_gazebo rviz.launch.py`

Confirm the wamv appears with all components drawn and in the right place.

`ros2 topic pub /wamv/thrusters/left/pos std_msgs/msg/Float64 "{data: -1}"`

Watch the left engine/propeller move in RViz as it moves in Gazebo.

`ros2 run tf2_tools view_frames`

Confirm the new TF tree is structured correctly.

I still haven't looked into the extra "wamv/" prepended on the PosePublisher's output, so this code adds the prefix to all ROS-generated link names instead, in order to make the sensor sdf links connect correctly.  This is minor but annoying, so if we decide to implement this I will take another look at it.


j-herman:
@M1chaelM @caguero 
This is ready for a look, but we'll need to discuss.  I do think it fixes the tf issues but I'm not sure if/when we want to make a significant change.  Marked both this and the other option (tutorial hack only) as draft pending decisions.

caguero (commented):
Everything works for me as expected, I only committed a very minor tweak.

My only suggestion is about the rViz configuration: should we include the three cameras and the lidar?

j-herman:
I don't think the extra cameras and lidar are necessary - it's super easy to add them, and one topic is shown as an example - but there's no reason not to have them all if you think that would look better for the demo.  Whichever you prefer works for me!

caguero (approved):

Issue #690, opened 2023-06-27
Gazebo crashing with violent WAM-V motion


M1chaelM:

caguero:
Updates:
  1. I managed to reproduce the crash without hitting any buoys. I did it spinning the WAM-V violently while braking and accelerating again causing some pitching.

M1chaelM:
While testing #692 I also managed to produce a crash in the acoustic tracking task world while moving relatively slowly. After a number of collisions the WAMV seems to destabilize and eventually tilts sideways under the water and crashes the simulation. A wild guess, but I wonder if this might also have something to do with the tuning of the drag coefficients, as described by @tejalbarnwal in #689 . The WAMV seems to be sitting high on the water and when the thrust is strong its movement seems strange.

caguero:
See a workaround patch in #706 .

caguero:
Fixed in #706

PR #706, opened 2023-07-15
Fix issue #690

caguero:
See #690 .

I tracked down the issue in the `SimpleHydrodynamics` plugin. If you create enough movement in the WAM-V, either spinning in circles or pitching up and down constantly, eventually, the plugin produces a very large combination of force and torque ending up with a crash in the physics `Step()` function.

To mitigate this issue I made the following changes:

1. I increased the damping of the hydrodynamics parameters.
2. I added an extra parameter and logic to limit the amount of force that the plugin applies. 

With these changes I haven't been able to produce a crash, although we have modified the WAM-V behavior a bit as a side effect.

Note: A better solution would be to understand why these high intensity forces are created when violent WAM-V movements are generated. I leave that for a future exercise because it requires a lot of debugging time.

**Update (July 19):** I don't think the problem is the hydrodynamics plugin, the problem is an abnormal velocity and acceleration in the vehicle. The hydrodynamics is just responding to that. I've decided to remove the cropping for this reason.

### How to test this PR?

Run a world and spin in circles in place. You normally will get a crash without this patch but the simulation should carry on with the patch in this PR.


M1chaelM:
It is definitely harder to produce a crash now, but I'm still able to do it. In addition, it seems that some of the interactions between the wamv and buoys has changed--the buoys appear to produce more resistance now when we run into them. I'm also able to flip them upside down. 

Here's a video of a crash with the current branch.

https://github.com/osrf/vrx/assets/8611855/03cb481d-667d-4183-aedc-4e0b12d963bd

Note that by spinning one direction and then the other, I still get a surprising acceleration, though it's harder to produce.

caguero:
@M1chaelM , I've tweaked a bit the parameters. Could you give it a shot please?

M1chaelM:
In the `sydney_regatta` world I'm still able to produce a crash, but it is more difficult to produce now, and the buoy behavior seems mostly back to normal. The WAM-V just moves a little slower, which is fine.

Unfortunately, the acoustic tracking task appears to be a very bad test case for this particular bug. I was not able to teleop through the task (doing my best to follow the beacon) without producing a crash. I did better than previously, but by the time I reached the second lap the wamv's movement started getting less stable, and it was rocking forward and back more than it should before it finally seemed to slip under the water and crash. 

Qualitatively, it seems like part of the issue is exacerbated by collisions, and part may have to do with moving the wamv in small bursts of acceleration, which produces accumulating pitch that eventually leads to a crash.

A couple notes towards tracking down the root of this problem. The crash outputs the following error message:
```
ODE INTERNAL ERROR 1: assertion "aabbBound >= dMinIntExact && aabbBound < dMaxIntExact" failed in collide() [collision_space.cpp:460]
```
This error is also referenced in the following open issue:
* https://github.com/gazebosim/gz-sim/issues/15
In addition, it looks like it may be related to this bug report in ardupilot that Rhys has already done a nice writeup of:
* https://github.com/ArduPilot/ardupilot_gazebo/issues/45

If I'm understanding correctly, Rhys's conclusion is that a change in the quadratic damping formula that gazebosim uses ultimately have the effect of causing an overflow in the angular velocity about x. The issue says the problem occurred with an upgrade to Gazebo 7.2, and Rhys appears to have a workaround for the rov in which the issue was reported:
* https://github.com/clydemcqueen/bluerov2_ignition/pull/5

I'm still using gazebo 7.0.0, but this sounds _a lot_ like what we're seeing, so I wonder if the change might have propagated down to 7.0.

caguero:
@M1chaelM , could you give it one more shot? I think we found the root of the problem and a fix. It shouldn't crash anymore.

caguero:
> @caguero With this update, I'm unable to move the WAM-V at all. Gazebo seems to be receiving teleop updates, so my guess is that the WAMV is staying locked.

Hmm, which world are you running?

M1chaelM:
> Hmm, which world are you running? And not after 10 seconds?

It was happening in both the `acoustic_tracking_task` world and the `sydney_regatta` world. I tried wiping the environment and building from scratch, but no luck. Changing back to the main branch restores the teleop capability. If there's a specific world I should try let me know.

caguero:
> > Hmm, which world are you running? And not after 10 seconds?
> 
> It was happening in both the `acoustic_tracking_task` world and the `sydney_regatta` world. I tried wiping the environment and building from scratch, but no luck. Changing back to the main branch restores the teleop capability. If there's a specific world I should try let me know.

Sorry, could you try again now? I think this was a problem related with the scale of the teleoperation config.

M1chaelM (approved):
Teleop did come back after that last change, and I was able to test. I ran `sydney_regatta` world and was unable to produce a crash. I was also able to complete `acoustic_tracking_task.sdf` without a crash. Before, I could only get half way through that task before I triggered a crash. Now I was able to do it without worrying about trying to move the WAMV carefully. It looks to me like the problem is gone and the WAMV behavior is now much more stable.

toorxx:
Hello @caguero and @M1chaelM,

I have been working with the WAM-V using VRX (Gazebo) along with OpenAI Gym to train a series of actions, and I came across the problem described in this PR. I implemented the proposed changes in my setup, but unfortunately, I am still facing the issue.

I am using the same version of Gazebo mentioned in this PR and have tested the WAM-V in the sydney_regatta world. Despite the WAM-V holding up for more training steps, I still experience the same error.

I wonder if you could consider reopening this case or provide me with additional guidance on how I might resolve this issue.
I am willing to collaborate on any further testing or in solving this problem if necessary. I greatly appreciate your help and your work on this project.

Thanks!

Issue #675, opened 2023-06-11
imu orientation data in not w.r.t standard frames 


aniket11bh:
**Issue**

`IMU` is giving orientation data w.r.t the initial frame of reference in which it is initialised when the sim starts.

**Expected behavior**
`IMU` should give data in `ENU` frame of reference.

**To Reproduce**
1. Write a ros2 node to convert imu data present on `/wamv/sensors/imu/imu/data` from quaternions to rpy and prints them on console. say the node name is imu_qt2rpy_publisher
2. Start the simulator
3. Start the imu_qt2rpy_publisher. 
4. Observe the data in console, yaw is around 0. (degrees)
5. Now, Try initialising the bot with a different orientation. By changing the yaw value to `0` in https://github.com/osrf/vrx/blob/main/vrx_gz/launch/competition.launch.py#L42
6. Observe the data from imu_qt2rpy_publisher, the `yaw` is still `0`.

> This means that the `IMU` is not giving data in standard frame of reference. This can cause issues later on while integrating `robot_localization` package.

**Possible fix:**
This can be achieved by adding following lines 
```xml
            <orientation_reference_frame>
              <localization>ENU</localization>
            </orientation_reference_frame>
```
under the `<imu>` tag at [wamv_imu.xacro#L33](https://github.com/osrf/vrx/blob/main/vrx_urdf/wamv_gazebo/urdf/components/wamv_imu.xacro#L33)

Fix ref : https://github.com/gazebosim/gz-sim/blob/gz-sim7/test/worlds/imu_heading_deg.sdf#LL85C1-L87C43 

**System Configuration:**
- Ubuntu 22.04
- ROS 2, Humble
- Gazebo Sim 7.0.0
- Nvidia GTX 650
- VRX

M1chaelM:
@caguero Could you take a look at this?

caguero:
> @caguero Could you take a look at this?

Sure, I'll take a look.

M1chaelM:
My understanding from discussion with @caguero  is that it is normal for an IMU to report values relative to its own frame of reference. However, @j-herman points out that this IMU is supposed to be part of an integrated GPS/IMU sensor, which is a bit different. We have changed the way we are simulating this, so we should be sure to check it's consistent with the behavior of VRX classic.

crvogt:
[...]

PR #708, opened 2023-07-21
Set IMU to reflect sensor position in global coordinates

crvogt:
In reference to Issue #675 

#### Issue:
IMU initializes to local reference frame on WAM-V spawn.
This can be checked by echoing IMU sensor data on spawn.

#### Fix Proposed by User:
Add the following to `wamv_imu.xacro` to link IMU to parent reference frame:
```
<orientation_reference_frame>
    <localization>ENU</localization>
</orientation_reference_frame>
```

#### Result
Launch with 
```
```
Confirm orientation values with 
```
ros2 topic echo /wamv/sensors/imu/imu/data
```
Change orientation values in `competition.launch.py` and re-launch. Changing spawn orientation to align with world axes shows IMU reflects orientation in global frame.

### DIFF
--- a/vrx_urdf/wamv_gazebo/urdf/components/wamv_imu.xacro
+++ b/vrx_urdf/wamv_gazebo/urdf/components/wamv_imu.xacro
@@ -101,6 +101,9 @@
               </noise>
             </z>
           </linear_acceleration>
+          <orientation_reference_frame>
+            <localization>ENU</localization>
+          </orientation_reference_frame>
         </imu>
       </sensor>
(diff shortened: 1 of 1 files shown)


caguero (approved):
It looks good to me. After the WAM-V stabilizes I get:

```
orientation:
  x: -0.0024969300736652216
  y: 0.0033636502299557915
  z: 0.4814398292664655
  w: 0.8764691335085422
```

which approximates to (0, 0, 1), which is the absolute orientation.

PR #709, opened 2023-07-25
Updated code to use angle as input for wavefield parameter direction

tejalbarnwal:
aims to address comments in #651 

Summary:
Previously, the direction 2D vector was normalized and then used as angles.
Updated code directly takes an angle as input and is used further.

Test:
the results are evident with `practice_2023_stationkeeping0_task.sdf`,  `practice_2023_stationkeeping1_task.sdf` and `practice_2023_stationkeeping2_task.sdf`. The direction is more evident when we view the worlds from the top view

Result:
`practice_2023_stationkeeping0_task.sdf`: mean wave direction angle=0.0

Screenshot from 2023-07-25 21-26-14

`practice_2023_stationkeeping1_task.sdf`: mean wave direction angle=45 degrees

Screenshot from 2023-07-25 21-27-11

`practice_2023_stationkeeping2_task.sdf`: mean wave direction angle=90 degrees

Screenshot from 2023-07-25 21-28-14

### DIFF
--- a/vrx_gz/src/Wavefield.cc
+++ b/vrx_gz/src/Wavefield.cc
@@ -58,7 +58,7 @@ class vrx::WavefieldPrivate
     amplitude(0.0),
     period(5.0),
     phase(0.0),
-    direction(1, 0),
+    direction(0.0),
     angularFrequency(2.0*M_PI),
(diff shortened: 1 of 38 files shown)


M1chaelM (approved):
Great--works as expected. When we redo the environmental parameters tutorial I will add a note indicating that the direction is an angle in radians.

Issue #693, opened 2023-06-28
Implement Competition Mode


M1chaelM:
During the competition we don't publish debugging topics and scoring, so as to discourage teams from building solutions that rely on this information during the tasks. In VRX classic we called this "competition_mode" and enabled it by setting a parameter called `non_competition_mode` to `false`. We need to implement this functionality in VRX 2.0+. It would also be nice to eliminate the double-negative by calling the parameter `competition_mode` and setting it to `true` when we want it to be on.

To do this the argument needs to be added to the `competition.launch.py` and then handled appropriately.

caguero:
Done in #710

PR #710, opened 2023-07-28
Create competition_mode argument

caguero:
Fix issue #693.

This patch allows us to pass an extra `competition_mode` command line argument to enable/disable certain ROS 2 topics. As an example, I disabled the wind debug topics when `competition_mode` is enabled.

### How to test it?

Run the simulation as usual:

```
ros2 launch vrx_gz competition.launch.py world:=practice_2023_stationkeeping2_task
```

Verify that the ROS 2 wind debug topics are available:

```
ros2 topic list

...
/vrx/debug/wind/direction
/vrx/debug/wind/speed
...
```

Now, stop the simulation and run it again in competition mode:
```
ros2 launch vrx_gz competition.launch.py world:=practice_2023_stationkeeping2_task competition_mode:=true
```

And verify again the ROS 2 debug wind topics:
```
ros2 topic list
```

You shouldn't see them now.

### DIFF
--- a/vrx_gz/launch/competition.launch.py
+++ b/vrx_gz/launch/competition.launch.py
@@ -33,6 +33,7 @@ def launch(context, *args, **kwargs):
     headless = LaunchConfiguration('headless').perform(context).lower() == 'true'
     robot_urdf = LaunchConfiguration('urdf').perform(context)
     gz_paused = LaunchConfiguration('paused').perform(context).lower() == 'true'
+    competition_mode = LaunchConfiguration('competition_mode').perform(context).lower() == 'true'
     extra_gz_args = LaunchConfiguration('extra_gz_args').perform(context)
 
     launch_processes = []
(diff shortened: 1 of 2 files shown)


M1chaelM:
The wind topics do not appear in competition mode, so that is working as expecting. I still am seeing some topics that I thought would not show up, like 
```
/vrx/stationkeeping/mean_pose_error
/vrx/stationkeeping/pose_error
```
Is this intended? Or is the plan to decide on the exact list of topics later on?

caguero:
> The wind topics do not appear in competition mode, so that is working as expected. I still am seeing some topics that I thought would not show up, like
> 
> ```
> /vrx/stationkeeping/mean_pose_error
> /vrx/stationkeeping/pose_error
> ```
> 
> Is this intended? Or is the plan to decide on the exact list of topics later on?

If we know already which topics need to be excluded I can include them now.

M1chaelM:
> If we know already which topics need to be excluded I can include them now.
OK, I thought we could start by looking at what was excluded by the previous version of this command. In gazebo classic the variable was called `non_competition_mode` and it was passed to `vrx.launch` [here's the line in VRX docker](https://github.com/osrf/vrx-docker/blob/398448824831b5cdc78a73afe6b6d467d7014fdf/vrx_server/vrx-server/run_vrx_trial.sh#L43).

Unfortunately I don't see any such variable implemented on the VRX side for any previous version of VRX, so now I'm wondering whether this has ever been functional.

caguero:
I just excluded some extra topics when vrx is in competition mode. [3553c48](https://github.com/osrf/vrx/pull/710/commits/3553c485923eb44c3abd250ed1f4a52de36179bc)

M1chaelM:
Checking competition mode for:
* [ ] Stationkeeping
* [ ] Wayfinding
* [ ] Gymkhana

Issue #722, opened 2023-08-17
Sensor update rates incorrect


ngxingyu:
**Describe the bug**
The technical guide states the update rate of imu is 100hz and gps is 20hz, but currently both are 15 in wamv_gazebo
https://github.com/osrf/vrx/blob/7aeb71d88dc36c2bc271488fc37bab3736cea09d/vrx_urdf/wamv_gazebo/urdf/components/wamv_gps.xacro#L4
https://github.com/osrf/vrx/blob/7aeb71d88dc36c2bc271488fc37bab3736cea09d/vrx_urdf/wamv_gazebo/urdf/components/wamv_imu.xacro#L3

caguero:
Thanks for reporting it, I confirm the issue. I'll create a fix soon.

caguero:
See https://github.com/osrf/vrx/pull/723.

caguero:
It should be fixed in #723 . Let us know otherwise. Thanks!

PR #723, opened 2023-08-17
Fix GPS/IMU update rates

caguero:
See issue #722 .

### How to test it?

Launch the simulation:

```
```

And verify the publication rates:
```
ros2 topic hz /wamv/sensors/gps/gps/fix
ros2 topic hz /wamv/sensors/imu/imu/data
```

They should be closed to 20Hz and 100Hz respectively. Take into account that if the real time factor is lower than `100%`, the values reported from `topic hz` should be proportionally lower.


M1chaelM (approved):
I think this means we're going to do a 3.2.2 release. @caguero Do you agree?

caguero:
> I think this means we're going to do a 3.2.2 release. @caguero Do you agree?

Indeed, this is an important fix to justify a 3.2.2 release.

PR #747, opened 2023-10-13
Initialize acoustic perception score

caguero:
This is a bug fix for the acoustic perception task. While running phase 2, I noticed an issue with the scoring. The score should be equal to the elapsed time needed to reach the pinger. However, if the team doesn't reach the pinger, the plugin was reporting a score of `0`. Keep in mind that the lower the score the better. Instead, in that case the score should equal to the total running time (e.g. `300`).

How to test it?

Run this task and let it timeout while watching the score.

```
ros2 launch vrx_gz competition.launch.py world:=practice_2023_acoustic_perception0_task
```

And then:

```
```

### DIFF
--- a/vrx_gz/src/AcousticPerceptionScoringPlugin.cc
+++ b/vrx_gz/src/AcousticPerceptionScoringPlugin.cc
@@ -86,8 +86,6 @@ void AcousticPerceptionScoringPlugin::Configure(const sim::Entity &_entity,
       gzerr << "Error creating visual marker" << std::endl;
     }
   }
-
-  this->SetScore(this->RemainingTime().count());
 }
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


vrx_gz/src/AcousticPerceptionScoringPlugin.cc
```diff
@@ -86,6 +86,8 @@ void AcousticPerceptionScoringPlugin::Configure(const sim::Entity &_entity,
       gzerr << "Error creating visual marker" << std::endl;
     }
   }
+
+  this->SetScore(this->remainingTime().count());
```
j-herman, line 90:
```suggestion
  this->SetScore(this->RemainingTime().count());
```

j-herman (changes requested):

j-herman:
@caguero This doesn't seem to be working as intended.  The score stays at zero when I run the acoustic perception task.
I didn't dig in too far, but it looks like SetScore isn't available until we get to the running state, and the remaining time stays at zero for the first couple of task message updates.  Maybe use the timeout score?

caguero:
> 6ac1771

Sorry, wrong code. It should work now.

j-herman (approved):
Works as described!

PR #748, opened 2023-10-13
Fix score in follow-the-path task

caguero:
While running phase 2 I noticed an issue in the scoring of the follow-the-path task.

If the course isn't completed, the score is overwritten with the timeout score, that was initialized to `200`. Now that the task grants points each time the vehicles crosses a gate, we shouldn't do anything special if the task timeouts. The score should be preserved.

How to test it?

Launch a follow-the-path task:

```
ros2 launch vrx_gz competition.launch.py world:=practice_2023_follow_path0_task
```

And then:

```
```

If you don't cross any gates, the score should be `0` instead of `200`.

### DIFF
--- a/vrx_gz/src/NavigationScoringPlugin.cc
+++ b/vrx_gz/src/NavigationScoringPlugin.cc
@@ -367,11 +367,6 @@ void NavigationScoringPlugin::Configure(const sim::Entity &_entity,
     return;
   }
 
-  // Set default score in case of timeout.
-  double timeoutScore = 200;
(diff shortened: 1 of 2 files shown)


j-herman:
I've reviewed this and it's working fine, but for some reason the button to submit the approval is hanging up.  I'll try again later, but no need for anyone else to test this.

caguero:
> I've reviewed this and it's working fine, but for some reason the button to submit the approval is hanging up. I'll try again later, but no need for anyone else to test this.

Thanks! Merging then.

PR #758, opened 2023-12-17
RoboBoat01 reference model

caguero:
This pull request adds 1 of 2 of the RoboBoat reference models. I decided to add it to this repository to make it easier to modify by RoboBoat users, as opposed to upstreaming it to Fuel.

* [x] Basic SDF.
* [x] Collisions based on simple-shape geometries.
* [x] Tweaked inertias
* [x] Thruster plugin
* [x] Hydrodynamics plugin
* [x] Surface plugin attached to each pontoon.

Screenshot from 2023-12-17 09-54-38

Screenshot from 2023-12-17 09-55-05

Screenshot from 2023-12-17 09-55-28

### How to test it?

Add the following block to `worlds/sydney_regatta.sdf`.

```
<!-- RoboBoat 01 -->
<include>
  <name>roboboat01</name>
  <pose>-528 184 0 0 0 0</pose>
  <uri>roboboat01</uri>
</include>
```

Recompile VRX and launch it:
```
```

Spin the thrusters:
```
gz topic -t /model/roboboat01/joint/right_engine_propeller_joint/cmd_thrust -m gz.msgs.Double -p 'data: 1'
```


j-herman (approved):
Loads fine and runs thrusters with no problems noted.
We may want to make some parameter adjustments in the future.  It seems like the vehicle is much more directionally stable than I'd expect.  Not a major issue as this is a simple example but if we could find some representative performance characteristics to try to match it would be nice to work on this.

PR #763, opened 2024-01-08
Update Surface.cc

MRChamp001:


caguero (commented):
Hi @MRChamp001 , could you explain the reason for this patch? Thanks!

PR #764, opened 2024-01-08
Add Configuration File Roboboat Teleop

crvogt:
### What it Does
Maps joysticks on gamepad to each thruster on roboboat resulting in a differential drive system. Scaling is decreased in comparison to WAM-V config.

### How to Test
See the [tutorial](https://github.com/osrf/vrx/wiki/teleop_roboboat_tutorial) for how it is used and testing. 

#### Side Note to Reviewer
I've had past issues with my gamepad axes not matching intended command direction. This config matches the WAM-V config for direction/axis, so should be fine, but verification is appreciated.

### DIFF
--- a/vrx_gz/config/roboboat.yaml
+++ b/vrx_gz/config/roboboat.yaml
@@ -1,0 +1,98 @@
+# Copyright 2022 Open Source Robotics Foundation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


caguero (changes requested):
I have a few comments about the tutorial:

1. I think we need to modify `nbpark.sdf` to include `roboboat01`. Could you add that part to the tutorial? It should be similar to what we do in:

https://github.com/osrf/vrx/wiki/running_roboboat_tutorial

2. Perhaps we could keep the default `wamv.yaml` configuration and instead, just use the `teleop_config` parameter when running the node:

```
ros2 launch vrx_gz usv_joy_teleop.py teleop_config:=/home/caguero/vrx2023_ws/src/vrx/vrx_gz/config/roboboat.yaml
```

3. Does your robot move smoothly when you teleoperate? Mine is very reactive...

caguero:
Friendly ping @crvogt .

caguero:
I went ahead and applied (1) and (2).

caguero (approved):

PR #788, opened 2024-10-07
Fixes to support building on jazzy

JayHerpin:
I haven't fully run everything yet, but this at least gets it building.

Presumably this should work on iron as well.


HDenial:
is this actually working?

HDenial:
it did build correctly but then says:
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught multiple exceptions when trying to load file of format [py]:
 - ModuleNotFoundError: No module named 'sdformat13'
 - InvalidFrontendLaunchFileError: The launch file may have a syntax error, or its format is unknown

HDenial:
A good lad on reedit was able to fix the launch:
 qTHqq:

    1.Install Gazebo from OSRF repos, not the ROS package vendors (the vendor packages DO seem to get installed with rosdep but I guess it's okay when the workspace is sourced?)

    2.Clone and modify JayHerpin's dev-jazzy branch to change the sdformat13 import to sdformat14 in vrx_gz/src/vrx_gz/model.py and vrx_gz/src/vrx_gz/payload_bridges.py

I'd definitely query about getting the Python bindings turned on in the vendor packages, though. This is a hassle.

I also think maybe in the interim you should chime in on Jay's PR and suggest that it not use the -vendor packages. Instead I think ticking up manually to the new version numbers would avoid auto-installing the -vendor packages when you run rosdep, and would help make sure there's not some unholy mix of the ROS-buildfarm-vendored and the OSRF-installed binaries that will cause headaches later

caguero:
Thanks for the patch. Unfortunately, this approach using ROS 2 Gazebo Vendor packages doesn't work because Python support is disabled. I implemented an alternative approach in #807 . You could give it a try if you have time.

Issue #753, opened 2023-11-06
Lidar detects water in ROS humble


tejalacc:
I switched to ros2. In ros2 , vrx repo, when i run ros2 launch vrx_gz competition.launch.py world:=sydney_regatta and visualize it on rviz , lidar detects sea water also alongwith added buoys. whereas this issue is not their in vrx classic.can you help me this.

caguero:
Thanks for reporting it. This is a well-known issue and something that we'll need to fix in the future.

Muscle-king:
I would like to ask, has the problem of returning lidar points from the sea surface not yet been solved?

caguero:
It hasn't been resolved yet.

adarsha-s-bharadwaj:
Yes, even I'm facing this same issue.

JonathanWheare:
Hi everyone,

I ran into this problem as well and have done a little investigation of my own.  It appears that the water model is tagged with a visibility filter bit, that means that the object could be filtered by a sensor.  Unfortunately, there is no corresponding visibility setting in the lidar sensor.
I added the visibility filter tag to the lidar, and I found the water was filtered successfully.  I am putting together a PR submission for this change.

Regards,
Jonathan.

Dashenboy:
Hey @JonathanWheare Any Luck?

JonathanWheare:
Thanks for the reminder.  I have submitted a pull request.

caguero:
Fixed in https://github.com/osrf/vrx/pull/796

PR #796, opened 2025-02-02
In the vrx simulation, the ocean is currently detected by the lidar. …

JonathanWheare:
… where it is expected that the majority of lidar hits are absorbed by water.  Generally only wakes and waves will cause hits.

This patch adds a visibility_filter tag to the 3d lidar that will filter water hits, more closely matching the expected behaviour in the field.  The corresponding tag is already present in the water mesh.


JonathanWheare:
This addresses issue #753

JMANY42 (approved):
I implemented this in my local code base and it worked perfectly (and fixed a longtime headache of mine). I'm a bit of a git novice so forgive me if I'm doing anything wrong or not following correct procedure, but I wish to approve this commit.

caguero (approved):
Thanks Jonathan!

PR #807, opened 2025-05-02
Update the project to use Ubuntu 24.04, ROS 2 Jazzy and Gazebo Harmonic

caguero:
Requires [this PR from HonuRobotics/dockwater](https://github.com/HonuRobotics/dockwater/pull/37).

Instructions to test:

1. From the `dockwater` repo:
```
DIST=jazzy
./build.bash ${DIST}
./run.bash dockwater:jazzy
```

2. Within the `Docker` instance:
```
mkdir -p ~/vrx_ws/src
cd ~/vrx_ws/src
git clone https://github.com/osrf/vrx.git -b harmonic_jazzy
cd ..
colcon build --merge-install
```

3. Test it:
```
. install/setup.bash
ros2 launch vrx_gz competition.launch.py world:=sydney_regatta
```

Screenshot from 2025-05-02 15-09-20

Note: I'll update the installation instructions when this patch is merged.

PR #809, opened 2025-05-08
Add Honu Robotics headers to copyrights

caguero:
Adding some missing copyrights entries after #807.


bsb808 (approved):
Awesome.  Thank you for taking care of that!

PR #810, opened 2025-05-09
Migrate CI to use Jazzy + Harmonic: use ros-tooling actions

j-rivero:
This is a full rewrite of the CI system mostly to use common artifacts in the ROS Community, summary of the changes:

 * Do not use a base image from npslearninglab anymore. Instead of use one of the OCI images created in https://github.com/sloretz/ros_oci_images that are updated more often than the official dockerhub, see https://discourse.ros.org/t/docker-images-for-ros-updated-more-frequently/36931
 * The CI now runs daily to check for new Jazzy packages added to the repository in the case that we could find regressions in updates.
 * It runs on every push to the repository
 * Use ros-tooling/setup-ros-ci github action to configure a ros environment and install some build tools not in the docker image
 * Use ros-tooling/action-ros-ci to handle the compilation and testing by providing a colcon-defaults configurations

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,13 +14,12 @@ jobs:
         ros_distribution: [jazzy]
         include:
           - ros_distribution: jazzy
-            ubuntu_version: 22.04
             # desktop-full provides less time spent in installing 179 pkgs with respect to desktop
             # in exchange for using 500Mb more.
             ros_container: ghcr.io/sloretz/ros:jazzy-desktop-full
 
     name: ${{ matrix.ros_distribution }}
-    runs-on: ubuntu-${{ matrix.ubuntu_version }}
+    runs-on: ubuntu-latest
     container: 
       image: ${{ matrix.ros_container }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,33 +1,54 @@
-name: Ubuntu CI
+name: ROS 2 CI
 
-on: [pull_request]
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 0 * * *'  # every day to test Jazzy packages updates
 
 jobs:
-  test-humble-garden:
+  build-and-test:
+    strategy:
+      fail-fast: false
+      matrix:
+        ros_distribution: [jazzy]
+        include:
+          - ros_distribution: jazzy
+            ubuntu_version: 22.04
+            # desktop-full provides less time spent in installing 179 pkgs with respect to desktop
+            # in exchange for using 500Mb more.
+            ros_container: ghcr.io/sloretz/ros:jazzy-desktop-full
 
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,33 +1,54 @@
...
 jobs:
-  test-humble-garden:
+  build-and-test:
+    strategy:
+      fail-fast: false
+      matrix:
+        ros_distribution: [jazzy]
+        include:
+          - ros_distribution: jazzy
+            ubuntu_version: 22.04
```
caguero, line 17: Should this be `24.04`. [We're saying](https://github.com/osrf/vrx/wiki/system_requirements) that this is our recommended Ubuntu version.

.github/workflows/ci.yml
```diff
@@ -1,33 +1,54 @@
...
+          required-ros-distributions: ${{ matrix.ros_distribution }}
 
-      - name: Build and run tests
-        shell: bash
-        run: |
-          cd /home/developer/vrx_ws;
-          su -s /bin/bash -c "source /opt/ros/humble/setup.bash && colcon build --merge-install --executor sequential" developer;
-          su -s /bin/bash -c "source /opt/ros/humble/setup.bash && colcon test --merge-install --executor sequential --packages-skip-regex ros_gz* --return-code-on-test-failure --event-handlers console_direct+" developer;
+      - name: Build and test
+        uses: ros-tooling/action-ros-ci@1ff2c804b4c2383146d8cd8444dfda64ab4bf7ac  # v 0.4.3
```
ahcorde, line 37: why you don't use the tag ? `0.4.3` ?

ahcorde (changes requested):

caguero (commented):
Where are we installing Gazebo Harmonic?

.github/workflows/ci.yml
j-rivero: It is a security protection that has been working fine in real situations were a malicious attack happen in the github repository and the attacker modifies the tag to point to malicious code. Faking the hash is far more complicate. More info https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

.github/workflows/ci.yml
j-rivero: Jazzy can only be installed in Ubuntu Noble and we are using the jazzy images produced by Shane, so yes, we were using Noble. The 22.04 refers to the runner-system, I removed it to make clearer in [f77d96b](https://github.com/osrf/vrx/pull/810/commits/f77d96b8cf22ac68aa0410c53ee806ab1443a499)

j-rivero:
> Where are we installing Gazebo Harmonic?

We are installing Gazebo Harmonic from ROS 2 Packages. Since there is no runtime test affecting python, I chose not to install Gazebo packages from packages.o.o since it provides a faster CI cycle. I think that could be good to add another action or change this one to install the packages.o.o repo and use latest .debs in there since it is the closed experience to the users and/or development that we are going to have.

caguero (approved):

PR #815, opened 2025-05-13
Import the dockwater/jazzy Dockerfile and cleanup

j-rivero:
Since the development Dockerfile used for the repository was referred to be the one in https://github.com/HonuRobotics/dockwater/tree/main/jazz the PR import it here to be close to the code and easy to find by the users.

The first commit imported the file as-it is in the dockwater repository and the second 05a270448181a5667da13039898630dfb5c83f70 does a full clean-up including:

 * Remove all sudo references
 * Re-order steps 
 * From the list of development packages, remove those already installed by gz-harmonic, ROS 2 Jazzy and the Ubuntu base image.

From here we can think on creating different docker images to be use in CI, dev and releasing based on the same Dockerfile structure.

### DIFF
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -1,0 +1,89 @@
+# Ubuntu 24.04 
+FROM osrf/ros:jazzy-desktop-full
+
+ARG ROSDIST=jazzy
+
+# Setup timezone
+ENV TZ=Etc/UTC
+RUN echo $TZ > /etc/timezone && \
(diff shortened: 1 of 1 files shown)


caguero (approved):
Works for me!

j-rivero:
Pushed the changes to the jazzy branch directly.

caguero:
> Pushed the changes to the jazzy branch directly.

FYI: I don't see your changes in `jazzy`.

PR #472, opened 2022-07-13
require wave_gazebo_plugins to use c++17

acxz:
Solves #471

Issue #471, opened 2022-07-13
plugins use features of c++17


acxz:
Portion of build log:
```
In file included from /usr/include/gazebo-11/gazebo/physics/physics.hh:9:
/usr/include/gazebo-11/gazebo/physics/Collision.hh:205:28: error: ‘optional’ in namespace ‘std’ does not name a template type
  205 |       public: virtual std::optional<sdf::SemanticPose> SDFSemanticPose()
      |                            ^~~~~~~~
/usr/include/gazebo-11/gazebo/physics/Collision.hh:205:23: note: ‘std::optional’ is only available from C++17 onwards
  205 |       public: virtual std::optional<sdf::SemanticPose> SDFSemanticPose()
      |                       ^~~
In file included from /usr/include/gazebo-11/gazebo/physics/physics.hh:26:
/usr/include/gazebo-11/gazebo/physics/Light.hh:78:20: error: ‘optional’ in namespace ‘std’ does not name a template type
   78 |       public: std::optional<sdf::SemanticPose> SDFSemanticPose() const override;
      |                    ^~~~~~~~
/usr/include/gazebo-11/gazebo/physics/Light.hh:78:15: note: ‘std::optional’ is only available from C++17 onwards
   78 |       public: std::optional<sdf::SemanticPose> SDFSemanticPose() const override;
      |               ^~~
In file included from /usr/include/gazebo-11/gazebo/physics/physics.hh:28:
/usr/include/gazebo-11/gazebo/physics/Link.hh:617:20: error: ‘optional’ in namespace ‘std’ does not name a template type
  617 |       public: std::optional<sdf::SemanticPose> SDFSemanticPose() const override;
      |                    ^~~~~~~~
/usr/include/gazebo-11/gazebo/physics/Link.hh:617:15: note: ‘std::optional’ is only available from C++17 onwards
  617 |       public: std::optional<sdf::SemanticPose> SDFSemanticPose() const override;
      |               ^~~
In file included from /usr/include/gazebo-11/gazebo/physics/physics.hh:36:
/usr/include/gazebo-11/gazebo/physics/PhysicsEngine.hh: In static member function ‘static T gazebo::physics::PhysicsEngine::any_cast(const boost::any&)’:
/usr/include/gazebo-11/gazebo/physics/PhysicsEngine.hh:274:45: error: ‘any’ is not a member of ‘std’
  274 |           auto value = boost::any_cast<std::any>(_value);
      |                                             ^~~
[...]

PR #828, opened 2025-06-19
Use C++17

caguero:
Solves #471 using C++20 and replace #472 targeting `jazzy`.


ahcorde (changes requested):
I think we should use version `17` which is the required version on `jazzy`. Version `20` will be used in L-turtle

caguero:
> I think we should use version 17 which is the required version on jazzy. Version 20 will be used in L-turtle

C++17, then. [26c7605](https://github.com/osrf/vrx/pull/828/commits/26c76059a0103a8058bbbe205c60c201a46b5551)

ahcorde (approved):

PR #840, opened 2025-07-17
Inject custom vendor packages for Harmonic to avoid Python ABI problems

### LINKED ISSUE #839: jazzy branch possibly broken
Hi!

First of all, slightly off-topic, it would be nice to have at least 1 test launching the system. Right now only compilation is tested and issues with vendor and not vendor packages are not. (from what i can understand)

My current issue trying to rebase on vrx/jazzy is that [](http://packages.osrfoundation.org/gazebo/ubuntu-stable) provides sdformat `14.8.0` and the installed vendor package comes with `14.7.0`.

I see [https://github.com/gazebo-release/sdformat_vendor/pull/15](https://github.com/gazebo-release/sdformat_vendor/pull/15) , which is not merged. This means that when trying to launch the sw stack i get incompatible symbols.

Summary:
- gz-harmonic shipped with (http://packages.osrfoundation.org/gazebo/ubuntu-stable) depends on sdformat, and installs python bindings for `14.8.0` .
- ros-jazzy-desktop-full now installs vendor at `14.7.0` via:
    ```
    ros-jazzy-desktop-full 
      └── ros-jazzy-ros-gz-sim-demos 
          └── ros-jazzy-ros-gz-sim 
              └── ros-jazzy-sdformat-vendor
    ```

There are [many deps on vendor on the ros-gz codebase](https://github.com/gazebosim/ros_gz/blob/5995c031fe06b75ad3a1304c7a88323609b5b119/ros_gz_sim/package.xml#L30-L33)... so I don't thinkg building from source on a custom branch is a valid solution here.

Thoughts? @caguero @j-rivero

j-rivero:
Closes: #839 

To avoid the installation of the ros-jazzy-gz-*-vendor (being * the Gz packages with Python wrappers)  / ros-jazzy-sdformat-vendor packages the PR does the following:

 * The PR changes the Dockerfile.base not to be based in desktop-full but in ros-base
 * Create a fake debian package that `Provides:` the vendor packages for the Gazebo libraries that have Python bindings. This way the .deb package dependency chain can continue working without the risk of installing the vendor packages and rosdep can still be used.
 * To replace the real ROS Gazebo vendor packages "provided" by the fake .deb package, stablish a colcon workspace with a gz.repos file that builds from source the packages. The Gazebo vendor packages can use the system installation of Gazebo libraries if present so no real compilation is done and the .deb packages are being used.
 * Inject for interactive and non-interactive logins the variables that allows the transparent interaction for the user sourcing the colcon workspace and making the `gz` tool from ROS to work also with system packages.

### DIFF
--- a/docker/Dockerfile.base
+++ b/docker/Dockerfile.base
@@ -1,5 +1,5 @@
-# Ubuntu 24.04 
-FROM osrf/ros:jazzy-desktop-full AS vrx-base
+# Ubuntu 24.04
+FROM ros:jazzy-ros-base AS vrx-base
 
 # Workaround from https://discourse.ros.org/t/ros-signing-key-migration-guide/43937
 RUN rm -f /etc/apt/sources.list.d/ros2-latest.list \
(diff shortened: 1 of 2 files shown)


caguero (commented):
I was able to reproduce the issue by running:
```
rocker --devices /dev/input/js0 --x11 --nvidia --user --home ghcr.io/osrf/vrx-devel:latest /bin/bash
```

and then inside the container
```
sudo apt-get update
sudo apt-get upgrade
```

After compiling VRX, I hit the issue when launching VRX:
```
caguero@1f7431d4c4df:~/vrx_ws$ ros2 launch vrx_gz competition.launch.py
[INFO] [launch]: All log files can be found below /home/caguero/.ros/log/2025-07-17-22-06-38-861309-1f7431d4c4df-5099
[INFO] [launch]: Default logging verbosity is set to INFO
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught multiple exceptions when trying to load file of format [py]:
 - ImportError: /usr/lib/python3/dist-packages/sdformat14.cpython-312-x86_64-linux-gnu.so: undefined symbol: _ZNK3sdf3v144Cone14AxisAlignedBoxEv
 - InvalidFrontendLaunchFileError: The launch file may have a syntax error, or its format is unknown
```

Do you have some instructions to verify that this patch fixes the issue?

j-rivero:
> Do you have some instructions to verify that this patch fixes the issue?

Oh yes, sorry.

```bash
cd docker
docker-compose build base devel
rocker --devices /dev/input/js0 --x11 --nvidia --user --home vrx-devel:latest /bin/bash
```

Should leave you in the modified container.

caguero (approved):
It works for me now!

PR #863, opened 2025-11-03
[kilted] bump gz lib versions to ionic

Rezenders:
These changes should actually go into a `kilted` branch, but since that is not available I'm creating this PR as a draft.
If you make a `kilted` branch available, we can convert this PR from draft.


caguero (changes requested):
Thanks!

Could you also grep for `jazzy` and update all the references? There are some Docker files to be updated as well?

caguero:
Superseded by #871. Please, let me know if you're interested in working on the Kilted branch.

PR #865, opened 2025-11-14
Set executable flag on generate_wamv.py script

rolker:
This fixes an `executable 'generate_wamv.py' not found` error when launching generate_wamv.launch.py


caguero (approved):
Thanks @rolker !

Issue #866, opened 2025-11-17
Joints oscilliations affect sensor readings (IMU).


stepanove:
Hello!

I just started playing with VRX simulation environment and, first of all, I'd like to thank you for your work, it's cool to have an opportunity to experiment with marine robotics.

I noticed an issue with some joints starting to oscillate if you "shake" WAM-V a bit. It's visible with a naked eye and also affects data readings, for example IMU. The ball shooter is the most prominent feature, but other axes are also shaking in RViz, it's a bit harder to see, though.

The issue affects following axes:
- wamv/ball_shooter_base_joint
- wamv/ball_shooter_launcher_joint
- wamv/gps_wamv_joint
- wamv/imu_wamv_joint
- wamv/receiver_pinger_joint

It looks like this happens because these axes have zero damping and friction, probably they should be damped or made fixed instead of revolute.

**To Reproduce**
1. Run  ```ros2 launch vrx_gz competition.launch.py```
2. Next, do some high-G maneuvers using gamepad or run ```ros2 topic pub -1 /wamv/thrusters/left/thrust std_msgs/msg/Float64 {"data: 1000."} && ros2 topic pub -1 /wamv/thrusters/right/thrust std_msgs/msg/Float64 {"data: 1000."} && ros2 topic pub -1 /wamv/thrusters/left/thrust std_msgs/msg/Float64 {"data: 0."} && ros2 topic pub -1 /wamv/thrusters/right/thrust std_msgs/msg/Float64 {"data: 0."}```
3. Repeat step 2 until the ball shooter begins to oscillate in Gazebo, one time is typically enough.

**System Configuration:**
- OS: vrx-devel:v3.0.4
- ROS Version: ROS 2, Jazzy
- Gazebo Version: 8.9/8.10, the behaviour is the same with the stock image which has Gazebo 8.9 and with Gazebo updated to 8.10.

https://github.com/user-attachments/assets/1c1518e8-548c-4c52-83cc-c629f32135dd

PR #867, opened 2025-11-18
Make IMU, GPS, pinger and ball shooter joints fixed

stepanove:
This is a proposed fix for the issue #866.
I'm new to VRX, but as far as I understand, the ball shooter is not supposed to be actuated. Actually, the current joints configuration is wrong anyway, it has two joints both rotating around Z, the second joint should rotate around Y, I suppose. It's obvious that IMU, GPS and acoustic sensors should be fixed. It looks like all of this is just a kind of a copy-paste mistake.
So my suggestion is to make all the following joints fixed:
- wamv/ball_shooter_base_joint
- wamv/ball_shooter_launcher_joint
- wamv/gps_wamv_joint
- wamv/imu_wamv_joint
- wamv/receiver_pinger_joint

### DIFF
--- a/vrx_urdf/wamv_gazebo/urdf/components/ball_shooter.xacro
+++ b/vrx_urdf/wamv_gazebo/urdf/components/ball_shooter.xacro
@@ -47,16 +47,12 @@
       </inertial>
     </link>
 
-    <joint name="${namespace}/${name}_base_joint" type="revolute">
-      <axis xyz="0 0 1"/>
-      <limit effort="1000.0" lower="0.0" upper="0" velocity="0"/>
+    <joint name="${namespace}/${name}_base_joint" type="fixed">
       <origin xyz="${x} ${y} ${z}" rpy="0 0 ${yaw}"/>
       <parent link="${namespace}/base_link"/>
       <child link="${namespace}/${name}_base_link"/>
     </joint>
-    <joint name="${namespace}/${name}_launcher_joint" type="revolute">
-      <axis xyz="0 0 1"/>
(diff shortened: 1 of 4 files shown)


stepanove:
`[gazebo-1] [Err] [BallShooterPlugin.cc:158] The frame 'wamv/ball_shooter_launcher_link' does not exist` 

I noticed an error from BallShooterPlugin, it turns out that Gazebo merges links with fixed joints by default. I pushed a fix to prevent it.

stepanove:
Hi @caguero , did you have a chance to take a look at this issue?
Is there some hidden reason in keeping those axes revolute that I'm missing? If the problem is only in the fixed joint lumping, then it would be better to use "preserveFixedJoint" option, I suppose.

caguero:
> Hi @caguero , did you have a chance to take a look at this issue? Is there some hidden reason in keeping those axes revolute that I'm missing? If the problem is only in the fixed joint lumping, then it would be better to use "preserveFixedJoint" option, I suppose.

I need to check old issues but I remember we had a problem with a fixed joint related with joint lumping and this was the workaround.

caguero:
> I need to check old issues but I remember we had a problem with a fixed joint related with joint lumping and this was the workaround.

Here's [one related issue](https://github.com/osrf/vrx/pull/669).

stepanove:
> > I need to check old issues but I remember we had a problem with a fixed joint related with joint lumping and this was the workaround.
> 
> Here's [one related issue](https://github.com/osrf/vrx/pull/669).

I see, thanks! I added "preserveFixedJoint" option to all joints I made fixed previously.

Issue #869, opened 2025-11-20
Allow external sensors to be added to WAMV


rolker:
Allow other sensors that the ones pre-defined, such as a multibeam sonar sensor.

This helps accomplish #868

PR #870, opened 2025-11-20
Allow external components

rolker:
This allows additional sensors to be added to the robot.  Addresses #869.


caguero (approved):
Thanks for the patch!

PR #880, opened 2026-06-02
Add GitHub Actions CI with a headless rendering smoke test

j-rivero:
## Add GitHub Actions CI with a headless rendering smoke test

Bootstraps continuous integration for `vrx4`. There was no CI before; this builds the workspace and verifies the simulation actually comes up and renders on every push and PR.

### CI workflow (`.github/workflows/ci.yml`)

A single `smoke` job that runs on `push`, `pull_request` against `vrx4`, and `workflow_dispatch`:

- Runs in the `ros:lyrical-ros-base` container.
- Installs **Gazebo Jetty** via the `ros-gz` packages (`ros-lyrical-ros-gz-sim`, `ros-lyrical-ros-gz-bridge`) from the `ros2-testing` repo, then resolves the remaining workspace deps with `rosdep`.
- Builds with `colcon build --merge-install`, and explicitly asserts that `vrx_bringup` was installed (colcon exits 0 even when it builds nothing).
- Caches the Gazebo Fuel assets pulled by the world so reruns are fast and less network-dependent (cached under `/github/home/.gz/fuel` to match the container `HOME`).

### Headless smoke test (`tools/smoke_test.sh` + `tools/check_render.py`)

Exercises the **real** production launch (`vrx_bringup simulation.launch.xml`, `gazebo_gui:=false`) under a virtual X server (Xvfb) with Mesa software GL (llvmpipe), so rendering is validated without a GPU. The flow:

1. Launch the production sim headless.
2. Wait for the sim to come up via rclpy `/clock` detection (`check_render.py --wait-only`).
3. Spawn a **test-only** observation camera into the running world and bridge its image to ROS just for the test.
4. Assert over a measurement window that `/clock` is advancing (physics is stepping, not paused) **and** the camera is publishing frames — reporting the achieved frame rate.

It isolates its ROS graph per run, fails fast on launch crashes, and dumps the launch log on any failure.

### Test-only observation camera

The observation camera is deliberately **not** part of any shipped world. It lives in `tools/observation_camera.sdf` and is spawned into the running sim at test time via `ros_gz_sim create` (the world already loads the Sensors system, so it renders). The production world, launch file, and bridge config carry no CI scaffolding.

### Verification

Green end-to-end in CI. The smoke step confirms the runtime-spawned camera renders under software GL:

```
/clock detected; simulation is up.
[smoke] spawning observation camera into world 'default'...
clock advancing: True (688000000 -> 4266000000 ns)
CAMERA FPS: 6.70 (67 frames on /observation_camera/image over 10.0s)
PASS: sim stepping and camera rendering.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,68 @@
+name: vrx4 CI
+
+on:
+  push:
+    branches: [jrivero/vrx4_ci_clean]
+  pull_request:
+    branches: [vrx4]
+  workflow_dispatch:
+
+jobs:
+  smoke:
+    name: Smoke xvfb (Lyrical + Jetty)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    container:
+      image: ros:lyrical-ros-base
+    env:
+      DEBIAN_FRONTEND: noninteractive
+    steps:
+      - name: Headless GL + ros2-testing repo + checkout deps
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends \
+            ca-certificates git \
+            xvfb libgl1-mesa-dri libglx-mesa0 \
+            mesa-utils \
+            ros2-testing-apt-source
+
+      - uses: actions/checkout@v4
+        with:
+          path: src/vrx
+
+      # open_water.sdf pulls the Portuguese Ledge tile from Gazebo Fuel on first
+      # run; cache it so reruns are fast and less network-dependent.
+      - name: Cache Gazebo Fuel assets
+        uses: actions/cache@v4
+        with:
+          # In a job container HOME is /github/home, so Gazebo writes its Fuel
+          # cache to /github/home/.gz (not /root/.gz).
+          path: /github/home/.gz/fuel
+          key: gz-fuel-${{ hashFiles('src/vrx/vrx_gazebo/worlds/*.sdf') }}
+          restore-keys: gz-fuel-
+
+      - name: Install ros-gz (Gazebo Jetty) + workspace deps
+        run: |
+          apt-get update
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+            ca-certificates git \
+            xvfb libgl1-mesa-dri libglx-mesa0 \
+            mesa-utils \
+            ros2-testing-apt-source
+
+      - uses: actions/checkout@v4
+        with:
+          path: src/vrx
+
+      # open_water.sdf pulls the Portuguese Ledge tile from Gazebo Fuel on first
```
caguero, line 33: Nit: I'd remove the concrete references to an sdf world or model, as this might change in the future.

tools/check_render.py
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env python3
```
caguero, line 1: Can you include a license header in this file?

tools/smoke_test.sh
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env bash
```
caguero, line 1: Should we add a license header?

tools/smoke_test.sh
```diff
@@ -0,0 +1,121 @@
...
+# Headless smoke test for vrx4: GLX rendering under a virtual X server (Xvfb)
```
caguero, line 2: Nit: I'd remove the reference to the specific branch and Gazebo/ROS versions, in case we change it in the future.

caguero (commented):
This code mostly feels generic. How do you feel about moving it to its own GitHub composite action? The VRX repo will stay cleaner and more projects can benefit from it.

PR #884, opened 2026-06-19
gz_waves: rendering - Part 3/4

caguero:
Third in the stacked series. Adds the renderer, the `WaterVisual` GUI system that draws the wave surface, plus a runnable Gerstner-ocean demo world.

## How to test it?

```bash
ros2 launch vrx_bringup simulation.launch.xml
```
## What this PR adds

- **`WaterVisual`** (`gz-sim-water-visual-system`) — a gz-sim GUI system that reads the `Wavefield` component, rebuilds a **thread-private** engine from the replicated recipe via the core registry (`CreateWaveSimulation`), and uploads its height/displacement grid as a GPU heightmap each frame. It is **provider-agnostic**: it links *no* wave engine and depends only on the `gz_waves` core.
- **`HeightMapTexture`** — the CPU→GPU heightmap helper.
- **`Ogre2HeightMapBridge`** (`libwaves-ogre2-bridge`) — all Ogre Next access is isolated behind a C-ABI in this separate library, loaded on demand via `dlopen`, so `WaterVisual` carries no `DT_NEEDED` on `libgz-rendering-ogre2` (which would interfere with gz-rendering's own engine-plugin loader).
- **`water_surface` model** (mesh + shaders + textures) — loads `WaterVisual` plus the per-engine **GUI registrar(s)** in the same `<visual>` (the only level the GUI's GuiRunner loads systems at). At this point that is `gz-sim-waves-gerstner-gui`; the FFT registrar line is added by PR4.
- **`vrx_gazebo/worlds/open_water.sdf`** — wired as a Gerstner ocean (gerstner source + `model://water_surface`, `<sea_state>` default + advanced-commented block) so `vrx_bringup`'s existing launch runs it directly. Ocean-only: no references outside the PR1–PR3 stack.

## Provider-agnostic design

The engine registration that lets `WaterVisual` rebuild an engine in the GUI process lives in the **per-engine GUI plugins** (shipped by the provider packages), not in this package. So adding or removing an engine never touches
`gz_waves_rendering` — it depends only on the core. `WaterVisual` keeps a private engine instance (not the component's) because the GUI render thread and the ECM thread both touch it; sharing would race.


gz_waves_rendering/src/systems/WaterVisual.cc
```diff
@@ -0,0 +1,777 @@
...
+  (*vsParams)["tau"] = this->cachedTau;
+
+  // Grid-path uniforms. The material always uses the grid vertex shader,
+  // so these bindings are always valid.
+  {
+    // The grid vertex shader uses world_matrix to compute world-space XY
+    // for the periodic heightmap sample, so tile instances at
+    // different world offsets each render their own piece of the
+    // continuous wavefield (rather than each tile showing the same
+    // patch in local model space). The Gerstner VS doesn't declare
```
bsb808: This comment says the Gerstner vertex shader doesn't declare `world_matrix`, which implies a second vertex shader exists. But this stack wires a single grid vertex shader (`fft_water_vs_330.glsl`) for every engine, and it *does* declare `world_matrix`. Is there still a separate Gerstner vertex shader, or is this comment vestigial from an earlier design where Gerstner had its own? If vestigial, the comment is misleading; if a second shader really exists, binding `world_matrix` unconditionally would throw on it — worth confirming which.

gz_waves_rendering/src/systems/Ogre2HeightMapBridge.cc
```diff
@@ -0,0 +1,341 @@
...
+      static_cast<Ogre::uint32>(_gridSize),
+      static_cast<Ogre::uint32>(_gridSize));
+  // Single mip: we only upload mip 0 each frame, and the engine doesn't
+  // auto-generate the rest (would need RenderToTexture + allowsAutoMipmaps,
+  // which conflict with our ManualTexture upload contract). With a full
+  // mip chain in place but only mip 0 populated, the FS would sample
+  // garbage from higher mips at distance — visible as fine dotted lines
+  // that drift with the waves. Matches `combinedTex` on the GPU path.
+  hm->texture->setNumMipmaps(1u);
+  // RGBA32F so we can pack (η, Dx, Dy, α) per texel. Alpha is unused for
```
bsb808: This comment appears to be stale? Says the alpha channel is "unused for now (reserved … for a future stage)," but the very next function packs the foam/folding metric into alpha (`waves_ogre2_heightmap_upload`) and the fragment shader reads it (the `useFoamMap` path). Suggest updating the comment to describe alpha as the folding/foam channel.

gz_waves_rendering/share/models/water_surface/shaders/water_fs_330.glsl
```diff
@@ -0,0 +1,145 @@
...
+// Visual extras (foam mask, Schlick Fresnel, reflection roughness,
+// crest colour modulation, multi-octave normals) live in follow-up
+// commits so each can be evaluated independently against this
+// baseline.
+
+#version 330
+
+uniform sampler2D bumpMap;
+uniform samplerCube cubeMap;
+// FFT heightmap. Only sampled when foamStrength > 0 (the gerstner
```
bsb808: This says the `heightMap` sampler "can be unbound" for the Gerstner path since `foamStrength=0`. But the *vertex* shader always samples `heightMap` for the displacement (the Gerstner engine produces a `Field()` grid too), so the sampler is never actually unbound — only the fragment-shader foam read is skipped. Suggest clarifying so it doesn't imply the texture is optional.

gz_waves_rendering/share/models/water_surface/shaders/water_fs_330.glsl
```diff
@@ -0,0 +1,145 @@
...
+
+  vec4 waterColor = mix(shallowColor, deepColor, facing);
+  vec4 color = mix(waterColor, envColor, waterEnvRatio);
+
+  // Foam overlay: blend toward white where the Tessendorf Jacobian
+  // identifies a compression zone.
+  float foam = ComputeFoamMask() * foamStrength;
+  color.rgb = mix(color.rgb, vec3(1.0), foam);
+
+  fragColor = vec4(color.xyz, 0.9);
```
bsb808: Water alpha is hardcoded to 0.9 (`vec4(color.xyz, 0.9)`). Is the surface intended to be translucent, and is the material's blend/depth state configured for alpha blending? A constant 0.9 alpha can cause depth-sorting artifacts against other transparent objects; if opacity isn't intended, 1.0 avoids the blend path entirely.  Should this be exposed as a user parameter?

bsb808:
CC and I reviewed.  Looking forward to the next PR with some full stack testing!
Looks really good.  Generalizing the engine and exposing the wave generation params is great.  Comments above are all just cleanup/clarification, so good-to-go.

Approved whenever @caguero is ready to merge.

gz_waves_rendering/src/systems/WaterVisual.cc
caguero: Vestigial from the old design where each engine had its own vertex shader; there is no separate Gerstner VS now, so I deleted that sentence in [dab473ec](https://github.com/osrf/vrx/commit/dab473ec85f392a916044919a282e57a4e52370f).

gz_waves_rendering/src/systems/Ogre2HeightMapBridge.cc
caguero: You are right, it is stale. Updated in [dab473ec](https://github.com/osrf/vrx/commit/dab473ec85f392a916044919a282e57a4e52370f) to describe alpha as the folding/foam channel packed in the upload path.

gz_waves_rendering/share/models/water_surface/shaders/water_fs_330.glsl
caguero: Clarified in [dab473ec](https://github.com/osrf/vrx/commit/dab473ec85f392a916044919a282e57a4e52370f): the texture is always bound (the vertex shader samples it for displacement); only the fragment shader foam read is gated by foamStrength.

gz_waves_rendering/share/models/water_surface/shaders/water_fs_330.glsl
caguero: Set to opaque 1.0 in [dab473ec](https://github.com/osrf/vrx/commit/dab473ec85f392a916044919a282e57a4e52370f) to keep the water out of the transparent path (no blend state is configured); we can expose a translucency parameter later if wanted.

PR #889, opened 2026-07-08
Bsb/lyrical jetty build

bsb808:
## Symptom

Building gz_waves_rendering on ROS 2 Lyrical + Gazebo Jetty fails at CMake configure:

```
CMake Error at CMakeLists.txt:48 (message):
  OgreRoot.h not found under /usr/include/OGRE-2.3 (probed from
  gz-rendering::ogre2 = _gz_rend_ogre2_lib-NOTFOUND).
```
The package never compiles; the FFT/Gerstner providers building in parallel get aborted with it.

## Explanation 

1. No consistent dev environment for the target stack. The wave packages target ROS 2 Lyrical + Gazebo Jetty

2. Brittle, hardcoded Ogre-Next header discovery. gz_waves_rendering/CMakeLists.txt locates the Ogre-Next headers by reading gz-rendering::ogre2's LOCATION property and regex-matching a /opt/ros/<distro>/ path, else falling back to a hardcoded /usr/include/OGRE-2.3. For an imported/interface target LOCATION is empty, so the regex never matches and it drops to the hardcoded fallback — which does not exist on Jetty (newer Ogre-Next).

3. Multiple OgreRoot.h in the environment. The Jetty container ships three, only one of which is Ogre-Next:

```
/opt/ros/lyrical/opt/rviz_ogre_vendor/include/OGRE/OgreRoot.h          # classic Ogre 1.x (rviz)
/usr/include/OGRE/OgreRoot.h                                           # classic Ogre 1.x (system)
/opt/ros/lyrical/opt/gz_ogre_next_vendor/include/OGRE-Next/OgreRoot.h  
```

A naive path guess can miss the right header or grab an incompatible classic-Ogre one — the detection only ever worked on the single environment it was written for.

## Proposed fix

a) Robust header discovery. Replace the LOCATION probe with a find_path for OgreRoot.h, seeded by the ogre2 target's own INTERFACE_INCLUDE_DIRECTORIES, with PATH_SUFFIXES limited to Ogre-Next dirs (OGRE-Next, OGRE-2.4, OGRE-2.3) so classic Ogre 1.x is never selected. Version-agnostic across Harmonic, Jetty, ROS-vendor, and system installs.

b) A Lyrical/Jetty dev container via Docker compose:

## Test / Verify
```
# 1. Build the Lyrical/Jetty dev image
cd ~/vrx_ws/src/vrx
./docker/build.bash docker lyrical            # -> vrx_dev:lyrical

# 2. Interactive dev container (GPU + X11 + home mount)
./docker/run_compose.bash                     # drops into bash in the container

# 3. Inside the container: build the wave stack
cd ~/vrx_ws
rm -rf build install log
colcon build --merge-install

# 4. Run the sim (EncinoWaves ocean, GPU-rendered)
source install/setup.bash
ros2 launch vrx_bringup simulation.launch.xml
```


docker/compose.yaml
```diff
@@ -0,0 +1,68 @@
...
+    #     prompt may read "I have no name!" — cosmetic, since this UID has no
+    #     /etc/passwd entry in the image. Drop this line to run as root. ---
+    user: "${MY_UID:-1000}:${MY_GID:-1000}"
+
+    # --- host networking: ROS 2 DDS discovery + simplest X11 (DISPLAY reaches
+    #     the host X server directly) ---
+    network_mode: host
+    ipc: host
+
+    working_dir: ${HOME}/WorkingCopies/oceansim/vrx_ws
```
caguero: This looks like a hardcoded path. Should we use `~/vrx_ws` instead?

docker/run.bash
```diff
@@ -0,0 +1,76 @@
...
+
+IMG_NAME=${@:$OPTIND:1}
+
+# Replace `:` with `_` to comply with docker container naming
+# And append `_runtime`
+CONTAINER_NAME="$(tr ':' '_' <<< "$IMG_NAME")_runtime"
+ROCKER_ARGS="${ROCKER_ARGS} --name $CONTAINER_NAME"
+echo "Using image <$IMG_NAME> to start container <$CONTAINER_NAME>"
+
+rocker ${CUDA} ${ROCKER_ARGS} $IMG_NAME bash
```
caguero: If we moved to compose because rocker does not know Ubuntu 26.04, does run.bash still work on Lyrical, or is it only meant for the other images?

docker/compose.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - DISPLAY=${DISPLAY}
+      - HOME=${HOME}
+      - QT_X11_NO_MITSHM=1
+      - NVIDIA_VISIBLE_DEVICES=all
+      - NVIDIA_DRIVER_CAPABILITIES=all
+
+    volumes:
+      - ${HOME}:${HOME}                                        # rocker --home
+      - /tmp/.X11-unix:/tmp/.X11-unix:rw                       # X11 socket
+      # No .Xauthority mount: `xhost +local:root` grants local access without it,
```
caguero: Small thing: this mentions xhost +local:root, but run_compose.bash uses xhost +SI:localuser. Could we make them consistent?

docker/compose.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+    environment:
+      - DISPLAY=${DISPLAY}
+      - HOME=${HOME}
+      - QT_X11_NO_MITSHM=1
+      - NVIDIA_VISIBLE_DEVICES=all
+      - NVIDIA_DRIVER_CAPABILITIES=all
+
+    volumes:
+      - ${HOME}:${HOME}                                        # rocker --home
```
caguero, line 52: Mounting the whole home looks quite broad. Just checking...

caguero:
In my case the sky doesn't render correctly within the container.

Screenshot from 2026-07-09 11-27-18

docker/build.bash
```diff
@@ -0,0 +1,45 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+
+# Builds a Docker image.
+# Usage: build.bash <dir containing Dockerfile> [tag]
+#   tag defaults to the resolved directory name, so passing '.' works too.
+image_name=vrx_dev
```
caguero, line 27: Is this image the same as the one referenced in `compose.yaml`? It says `vrx_dev:lyrical` in the latter.

caguero:
> In my case the sky doesn't render correctly within the container.

It should be fixed in 9341308

docker/compose.yaml
bsb808: Fixed in b315bccafe9f6bbd64b3985bd5a670a67a76cbc2

docker/run.bash
bsb808: Yes - rocker adds a dependency and seems to often lag.  removed this file

docker/compose.yaml
bsb808: That is my SOP.  I like it for dev containers.   It picks up .bashrc which can be convenient.

docker/compose.yaml
bsb808: modified the comment for consistency

docker/build.bash
bsb808: Added comment to clarify

gz_waves_rendering/CMakeLists.txt
```diff
@@ -28,27 +28,28 @@ find_package(gz-rendering REQUIRED COMPONENTS ogre2)
...
+# rviz_ogre_vendor), whose ABI-incompatible headers would break the build.
+get_target_property(_ogre2_incs gz-rendering::ogre2 INTERFACE_INCLUDE_DIRECTORIES)
+find_path(OGRE_NEXT_INCLUDE_DIR
+  NAMES OgreRoot.h
+  HINTS ${_ogre2_incs}
+  PATHS
+    /opt/ros/$ENV{ROS_DISTRO}/opt/gz_ogre_next_vendor/include
+    /usr/include
+  PATH_SUFFIXES OGRE-Next OGRE-2.4 OGRE-2.3
+)
```
caguero, line 65: `find_path` checks the standard system paths before these `PATHS`, so on a host with both `libogre-next-dev` and `gz_ogre_next_vendor` it can pick the mismatched `/usr/include` copy and silently reintroduce the slow Hlms path. Consider a first pass with NO_DEFAULT_PATH so the vendor headers win, then a plain fallback.

gz_waves_rendering/CMakeLists.txt
```diff
@@ -28,27 +28,28 @@ find_package(gz-rendering REQUIRED COMPONENTS ogre2)
...
+if(NOT OGRE_NEXT_INCLUDE_DIR)
```
caguero, line 66: `find_path` reuses a stale cached value even if the path no longer exists, so keeping the old EXISTS check on `OgreRoot.h` would still catch that case.

docker/compose.yaml
```diff
@@ -0,0 +1,65 @@
+# Dev container for VRX 4 (ROS 2 Lyrical + Gazebo Jetty), via plain Docker
+# Compose so it does NOT depend on rocker knowing Ubuntu 26.04.
+#
+# Mirrors:  rocker --home --user --x11 --nvidia --devices /dev/input/js0 ... bash
+#
+# Easiest:  ./docker/run_compose.bash
+#
+# Manually:
+#   xhost +SI:localuser:$(id -un)    # let the container reach your X server (once per login)
+#   cd ~/WorkingCopies/oceansim/vrx_ws/src/vrx
```
caguero: This personal path is still here. Maybe something generic like `cd <your vrx_ws>/src/vrx`.

caguero (commented):
A few more comments inline. It would also be nice to eventually add a CI job that builds this image and compiles gz_waves_rendering inside it, so Lyrical and Jetty breakage shows up early.

docker/run_compose.bash
```diff
@@ -0,0 +1,45 @@
...
+export MY_UID MY_GID
+MY_UID="$(id -u)"
+MY_GID="$(id -g)"
+
+# GPU selection: prefer the NVIDIA dGPU when one is present and usable on the
+# host, otherwise fall back to the Intel/AMD iGPU (base compose only). The
+# NVIDIA reservation refuses to start the container on a host without an NVIDIA
+# GPU, so only layer the overlay in when nvidia-smi actually reports one.
+compose_files=(-f "${SCRIPT_DIR}/compose.yaml")
+if command -v nvidia-smi >/dev/null 2>&1 && nvidia-smi -L >/dev/null 2>&1; then
```
caguero: If the driver is present but the nvidia container toolkit is not, compose fails to start instead of falling back to the iGPU. Maybe also check `docker info` for the nvidia runtime.

docker/run_compose.bash
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
caguero, line 1: `build.bash` carries the Apache header but this script does not. Should we add it for consistentcy?

docker/build.bash
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+# Tag: explicit 2nd arg, else the resolved directory basename (handles '.').
+distro="${2:-$(basename "$(cd "${context}" && pwd)")}"
+
+image_plus_tag=$image_name:$(export LC_ALL=C; date +%Y_%m_%d_%H%M)
+docker build --rm -t "$image_plus_tag" -f "${context}/Dockerfile" "${context}" && \
+docker tag "$image_plus_tag" "$image_name:$distro" && \
+echo "Built $image_plus_tag and tagged as $image_name:$distro" && \
+echo "To run:  ./docker/run_compose.bash"
```
caguero: Missing newline at end of file.

docker/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+      ros-${ROS_DISTRO}-ros-gz \
+      "ros-${ROS_DISTRO}-gz-*-vendor" \
+      ros-${ROS_DISTRO}-sdformat-vendor \
+ && rm -rf /var/lib/apt/lists/*
+
+# ---- Standard ROS packages + build tooling VRX uses -----------------------
+# NOTE: nav2 (navigation2, nav2-bringup, nav2-minimal-*) is intentionally
+# omitted — it is often not yet released early in a new distro's life. Add
+# ros-${ROS_DISTRO}-navigation2 etc. once available on Lyrical.
+RUN apt-get update \
```
caguero, line 55: This list can drift from our `package.xml` files over time. Maybe generate it with rosdep or note rosdep as the source of truth.

docker/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+      python3-rosdep \
+      python3-vcstool \
+ && rm -rf /var/lib/apt/lists/*
+
+# ---- Dev tools + NVIDIA GL (rocker --nvidia) ------------------------------
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends \
+      atop expect gdb gnutls-bin iputils-ping libbluetooth-dev libcwiid-dev \
+      net-tools python3-dbg python3-pip python3-venv vim wget xvfb \
+ && dpkg --add-architecture i386 \
```
caguero: Do we need the i386 GL libs? Nothing in VRX is 32 bit, so this block could probably go.

docker/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+      libxext6 libxext6:i386 libx11-6 libx11-6:i386 libglvnd0 libglvnd0:i386 \
+      libgl1 libgl1:i386 libglx0 libglx0:i386 libegl1 libegl1:i386 \
+      libgles2 libgles2:i386 \
+ && rm -rf /var/lib/apt/lists/*
+
+# ---- Source ROS for interactive + non-interactive shells ------------------
+RUN echo "source /opt/ros/${ROS_DISTRO}/setup.bash" > /etc/profile.d/ros.sh \
+ && chmod +x /etc/profile.d/ros.sh \
+ && echo "source /etc/profile.d/ros.sh" >> /etc/bash.bashrc
+ENV BASH_ENV=/etc/profile.d/ros.sh
```
caguero, line 92: Heads up: BASH_ENV makes every noninteractive bash source the ROS setup, which can surprise scripts that use set -u.

docker/Dockerfile
bsb808: The image is built before the workspace exists, so I don't think a build-time rosdep install fits without restructuring. Dependencies are resolved by rosdep against the mounted `~/vrx_ws` at build time. I've added a comment naming `package.xml` + rosdep as the source of truth and framing this apt list as a cache-priming convenience. 

For development, I prefer to have the source locally on the host and mount the host from the container.  Happy to switch to a rosdep-driven stage (with the source `COPY`'d in) if you'd prefer the image carry it. 29f91697

docker/Dockerfile
bsb808: Agreed — removed the i386 architecture, its extra `apt-get update`, and all `:i386` GL libs (a carryover from the desktop nvidia-docker pattern). Only the 64-bit GL/X11 libs remain. 949dee2f

docker/Dockerfile
bsb808: Intentional — `BASH_ENV` is what gives non-interactive `run_compose.bash <cmd>` (and `docker exec ... bash -c`) a sourced ROS environment. I've added a comment flagging the `set -u` interaction and the localized `set +u` workaround for any script that trips on it. 2dbbb12a

docker/build.bash
bsb808: Good catch — added the trailing newline. 983e0e00

docker/compose.yaml
bsb808: Replaced the personal path with a generic `cd <your vrx_ws>/src/vrx`. 983e0e00

docker/run_compose.bash
bsb808: Added the Apache-2.0 header to match build.bash (OSRF / 2026). 983e0e00

docker/run_compose.bash
bsb808: Good call — `nvidia-smi` only proves the host driver. Added a `have_nvidia_runtime()` check via `docker info` so the overlay is layered in only when Docker's nvidia runtime (from the container toolkit) is present; otherwise it falls back to the iGPU instead of failing. 50b80710

gz_waves_rendering/CMakeLists.txt
bsb808: Good call. Reworked to a `NO_DEFAULT_PATH` first pass so the vendor/target-hinted headers win, with a plain system fallback — no more picking a mismatched `/usr/include/OGRE-2.x`. b64bd63c

gz_waves_rendering/CMakeLists.txt
bsb808: Re-added an `EXISTS` guard on `OgreRoot.h` that unsets the cache when the path is stale, so it re-searches on the next configure. b64bd63c

caguero (approved):
Looks good!

Issue #93, opened 2019-05-23
Determine method for specifying the 'seastate' of the simulation.


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/vrx-gh-pages/#!/osrf/vrx/issues/93)) by Brian Bingham (Bitbucket: [brian_bingham](https://bitbucket.org/%7B9cdea257-248f-486e-8858-49d8485a916f%7D/)).**

----------------------------------------
For the purposes of the competition we need to be able to specify the wave conditions \(seastate\) in the environment - specifically we need to be able to give a range of seastate values to specify the range of operating conditions that the tasks solutions will be tested against.

In the current default branch the wave field is specified as a summation of N regular waves where the amplitude, period and direction of each wave component are specified directly - see [https://github.com/osrf/vrx/blob/master/wamv\_gazebo/urdf/dynamics/wamv\_gazebo\_dynamics\_plugin.xacro](https://github.com/osrf/vrx/blob/master/wamv_gazebo/urdf/dynamics/wamv_gazebo_dynamics_plugin.xacro)

In the proposed synchronized physics visualization \(see PR [#78 ](https://osrf-migration.github.io/vrx-gh-pages/#!/osrf/vrx/pull-requests/78/issue-23-coordinate-the-physics-and/diff)\) the wave field is specified like so…

```
./devel/lib/asv_wave_sim_gazebo_plugins/WaveMsgPublisher \
  --number 3 \
  --scale 2 \
  --angle 0.4 \
  --direction -1 -1 \
  --steepness 1 \
  --period 7 \
  --amplitude .1
```

where the properties of the individual wave components are determined as described [here](https://developer.nvidia.com/gpugems/GPUGems/gpugems_ch01.html). 

After we integrate the physics/visual synchronization \(see issue [#23](https://osrf-migration.github.io/vrx-gh-pages/#!/osrf/vrx/issues/23/coordinate-the-physics-and-visualization) (#23)\) we need to determine a reasonable way to describe the range of seastates that will be use for task evaluation.

PR #890, opened 2026-08-14
asv: add MAVLink actuator safety gateway

dc-tw:
## Summary

Adds an experimental WAM-V MAVLink actuator safety boundary without replacing
VRX's existing Gazebo thruster systems or MAVROS transport stack.

- Reuses MAVROS for MAVLink UDP, framing, CRC, and connection state.
- Adds a VRX-specific gateway with a 0.2 s actuator-command freshness watchdog,
  latched zero-thrust fault state, explicit reset, and controlled re-arm flow.
- Adds an independent MAVLink #93 producer, native Gazebo Transport setpoint
  monitor, stock-vs-gateway A/B harness, and transparent CSV analysis.
- Adds a Docker workflow for Ubuntu 24.04 / ROS 2 Jazzy / Gazebo Harmonic.
- Vendors only the unchanged upstream MAVROS 2.14 `hil.cpp` plugin with minimal
  build metadata because the Jazzy binary MAVROS package does not register it.

## Why

Stock VRX thrusters retain their most recent force setpoint. A producer can
therefore stop publishing while nonzero thrust remains commanded. This gateway
also covers the realistic case where MAVLink heartbeat remains healthy while
the actuator-control thread has stopped.

## Validation completed

- `asv_safety_scripts/run_unit_tests.sh` — 4/4 safety-policy tests pass.
- MAVROS 2.14 core plus the HIL overlay built successfully in a compatibility
  Docker check.

## Review limitation: native benchmark pending

Native Ubuntu 24.04 x86_64 latency validation has **not** been performed because
the current Apple Silicon development host under amd64 emulation is not a
representative safety-timing environment. The headless repeated-run harness is
included so this can be reproduced on the target platform:

```bash
./asv_safety_scripts/run_repeated_gateway_latency_test.sh --runs 50
```

Pending measurements are headless VRX A/B results for `T_zero_wall`, P95/max
latency, and dual-thruster `I_stale`. No unmeasured latency value is presented
as an empirical result, and Apple Silicon/QEMU timing is explicitly excluded
from safety-latency claims. Maintainer feedback and/or native x86_64 execution
of the supplied harness would be welcome.


dc-tw:
I am narrowing this change after review. The current PR mixes the runtime gateway, assignment-specific validation tooling, Docker workflow, and a vendored MAVROS HIL overlay. I am moving the gateway plus focused unit tests to a new, smaller PR; the validation harness remains in the assignment submission, and the MAVROS HIL packaging gap will be tracked upstream in MAVROS. Marking this PR as draft while the replacement is prepared.

dc-tw:
The focused replacement is now available as #891. It contains only the actuator safety gateway, MAVROS configuration, package metadata, and unit tests (415 added lines); assignment-specific Docker, producers, monitors, benchmark tooling, reports, and vendored MAVROS source are excluded. During the split I also confirmed that Jazzy officially packages the HIL plugin in `ros-jazzy-mavros-extras`, so the replacement declares `mavros_extras` as a dependency instead of opening an incorrect missing-plugin issue or carrying an overlay. Keeping this PR in Draft as the historical full-assignment prototype.

